PR #2, opened 2024-07-17
Use macros for tuple implementations of various types

luca-della-vedova:
### Implemented feature

As noted in the description. Formatting is quite opinionated but I couldn't run `cargo fmt` since it seems the codebase is not formatted and it would have blown up the diff.

Bevy provides an [all_tuples! macro](https://docs.rs/bevy/latest/bevy/utils/macro.all_tuples.html) that facilitates implementing macros for different sizes of tuples. This together with a macro implementation for the traits themselves makes it possible to reduce code duplication when implementing traits for tuples of different sizes.

This has been used for `Buffered`, `Bufferable`, `ForkCloneBuilder`, `Unzippable`, `UnzipBuilder` and `StreamPack`.


src/stream.rs
```diff
@@ -562,204 +563,166 @@ impl<T1: StreamPack> StreamPack for (T1,) {
...
+                        $(
+                            $T.0,
+                        )*
+                    ),
+                    (
+                        $(
+                            $T.1,
+                        )*
+                    )
+                )
```
luca-della-vedova, line 604: This is a bit of a symbol soup that I didn't know how to format better, it takes a lot of lines but it would be pretty much unreadable if it was compressed in a single line

src/chain/unzip.rs
```diff
@@ -29,174 +29,82 @@ pub trait Unzippable: Sized {
...
+                let mut targets = SmallVec::new();
+                let result =
+                    (
+                        $(
+                            {
+                                // Variable is only used to make sure this cycle is repeated once
+                                // for each instance of the $T type, but the type itself is not
+                                // used.
+                                #[allow(unused)]
+                                let $T = std::marker::PhantomData::<$T>;
```
luca-della-vedova, line 55: This is a "hack" I had to create to allow iterating over all the types passed as an input without actually using them

src/chain/unzip.rs
```diff
@@ -29,174 +29,82 @@ pub trait Unzippable: Sized {
...
+            fn distribute_values(
+                OperationRequest { source, world, roster }: OperationRequest,
+            ) -> OperationResult {
+                let Input { session, data: inputs } = world
+                    .get_entity_mut(source).or_broken()?
+                    .take_input::<Self>()?;
+
+                // Targets is cloned to avoid borrow checker issues when
+                // doing a mutable borrow of the world later
+                let targets = world.get::<ForkTargetStorage>(source).or_broken()?.clone();
```
luca-della-vedova, line 80: From my understanding previously the clone wasn't necessary because we would dereference the borrow to `Entity` which, because `Entity` implements `Copy`, would just copy it below the scenes.
Performance wise I believe cloning the `SmallVec` should be _almost_ the same as copying all the elements one by one

luca-della-vedova:
I found a trick using `PhantomData` and some minor refactoring that removed the limitations, not all the types have macro implementations.

src/chain/unzip.rs
mxgrey: Right, the problem here is we can't guarantee that the preallocated `SmallVec` size is big enough since it's currently [8](https://github.com/open-rmf/bevy_impulse/blob/main/src/operation.rs#L166), and these tuples go up to 12. At that point doing a clone will cause an avoidable heap allocation.

If there's any way to use a pattern here to create `N` variables with entities or to create a tuple of `N` entities, copied out of the `SmallVec`, that would be preferable. Heap allocations once in a while aren't a huge deal if avoiding them is too onerous, but the more we can avoid them inside of hot loops, the more compelling this library is.

luca-della-vedova: Fair, I "found" a trick that by adding a dummy macro element (that shouldn't really cost anything since it's only a compile time cost) we can unpack all the vector elements into a tuple, which also fairly simplifies the code there https://github.com/open-rmf/bevy_impulse/pull/2/commits/dc95e402ded1e812b0132ece141e038a7b0a3103.
This comes at the cost of a dependency on `itertools` since it has a very convenient method to turn an iterator into a tuple.

I had to change the maximum tuple size down to 12 since that's the maximum that itertools `next_tuple()` implements.

mxgrey: Awesome!

> I had to change the maximum tuple size down to 12 since that's the maximum that itertools next_tuple() implements.

This should be totally fine. Since it supports nested tuples of any depth, there's effectively no limit, but if people are exceeding tuples of 12 elements in one output then they're going pretty wild anyway.

src/chain/fork_clone_builder.rs
```diff
@@ -30,65 +32,57 @@ pub trait ForkCloneBuilder<Response> {
...
-
-        let u_0 = (self.0)(Chain::new(target_0, builder));
-        let u_1 = (self.1)(Chain::new(target_1, builder));
-        let u_2 = (self.2)(Chain::new(target_2, builder));
-        (u_0, u_1, u_2)
-    }
-}
+// Implements the `ForkCloneBUilder` trait for all tuples between size 2 and 15
+// (inclusive)
+all_tuples!(impl_forkclonebuilder_for_tuple, 2, 15, F, U);
```
mxgrey, line 88: Very minor nitpick, but let's cap the tuple size to 12 for all of them for the sake of consistency, so users don't get additionally confused by hitting different tuple limits for different functions. If someone wants more than 12 forked clones there are several ways they could still accomplish that with a limit of 12.

luca-della-vedova: https://github.com/open-rmf/bevy_impulse/pull/2/commits/903663c7d31ef50da5b69198011cd001c3e3dad7 also found that there was a tuple with size 1 implemented manually and removed that to make the diff even more red https://github.com/open-rmf/bevy_impulse/pull/2/commits/c97a7fcd82c5a355c44514616a89399c916c65c3

mxgrey (approved):
This is awesome!! Thanks for figuring out how to cook up the right symbol soup for all these traits :+1:

Issue #9, opened 2024-07-31
Provide a Plugin for downstream users


luca-della-vedova:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

Currently it is not terribly clear how users can use this library in their projects.
Some digging shows that the testing module has an implementation of "minimal plugins to make bevy_impulse work" [here](https://github.com/open-rmf/bevy_impulse/blob/e78cf8c2e3aa445d1629ba8e6416eb23583d260d/src/testing.rs#L46-L53) but doing the same doesn't actually work, specifically this is what I encountered when adding the system to an app in a different crate:

```
error: type `service::internal::ServiceHook` is private
   --> rmf_site_editor/src/lib.rs:219:9
    |
219 | /         app.insert_resource(DirectionalLightShadowMap { size: 2048 })
220 | |             .add_state::<AppState>()
221 | |             .add_systems(Update, bevy_impulse::flush_impulses)
    | |______________________________________________________________^ private type

warning: `rmf_site_editor` (lib) generated 9 warnings
error: could not compile `rmf_site_editor` (lib) due to 1 previous error; 9 warnings emitted
```

What we should probably be doing instead, is creating and exporting a public plugin that users can just add to their app through a `add_plugins` call.

mxgrey:
Right, I hadn't firmly decided on whether `flush_impulses` should be public or whether we should provide a plugin as the public API. IIRC the main problem with making `flush_impulses` public is that it has some system parameter which aren't public, and Rust will give a warning if you have a public function that leaks private members.

luca-della-vedova:
It seems for downstream crates having a non public type is a error, not a warning (i.e. the log I attached in the ticket), so it seems the crate can't currently be integrated in downstream projects

mxgrey:
I think this can be closed with the addition of https://github.com/open-rmf/bevy_impulse/commit/9686950650982b4268015e647c3cc938f2734ae6

PR #10, opened 2024-07-31
Migrate to Bevy 0.12

luca-della-vedova:
### Implemented feature

[rmf_site](https://github.com/open-rmf/rmf_site/blob/main/Cargo.toml) is currently on Bevy 0.12, so to use this library we would need to be at least at that version.

The bulk of the changes is `EntityMut` -> `EntityWorldMut`.
The only other change is in labels implementation, the test code shows a trait implementation example with a TODO that it would be a lot easier for users to derive it, but that would require refactoring this library to have a separate proc_macro crate.

Still pending #9 to have the crate actually usable but thought I wouldn't make the PR here too complex

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ authors = ["Grey <mxgrey@intrinsic.ai>"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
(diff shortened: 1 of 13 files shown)


mxgrey:
I wound up merging your changes manually, so I believe this PR can be closed. Let me know if I'm overlooking anything.

PR #11, opened 2024-08-02
Add a service spawning methods to World

luca-della-vedova:
### Implemented feature

Workflows can be spawned from `World` but services couldn't because of a missing implementation.

Added an implementation to spawn services from `World`, useful when `&mut World` access is available (i.e. when initializing resources)

### DIFF
--- a/src/service.rs
+++ b/src/service.rs
@@ -18,7 +18,7 @@
 use crate::{StreamPack, AddOperation, OperateService, Provider, ProvideOnce};
 
 use bevy::{
-    prelude::{Entity, App, Commands, Component, Deref, DerefMut},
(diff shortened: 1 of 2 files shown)


mxgrey (changes requested):
I think we can simplify this PR by just doing

```
impl<...> SpawnServicesExt<..> for World {
    fn spawn_service<...>(...) {
        let mut command_queue = CommandQueue::default();
        let mut commands = Commands::new(&mut command_queue, self);
        let provider = commands.spawn_service(builder);
        command_queue.apply(world);
        provider
    }
}
```

then we don't need to make any changes to `service_builder.rs`.

mxgrey (approved):
I've pushed the changes that I recommended

PR #27, opened 2024-10-02
Feature: Dynamic Workflows

koonpeng:
### Implemented feature

Add support for dynamic workflows, dynamic workflows are workflows that is built at runtime.

### Next steps

Add deserialization of dynamic workflows.


diagram.schema.json
```diff
@@ -0,0 +1,240 @@
...
+            "type": {
+              "type": "string",
+              "enum": [
+                "terminate"
+              ]
+            }
+          }
+        },
+        {
+          "description": "Connects the request to a registered node.\n\n``` # bevy_impulse::Diagram::from_json_str(r#\" { \"ops\": { \"start\": { \"type\": \"start\", \"next\": \"nodeOp\" }, \"nodeOp\": { \"type\": \"node\", \"nodeId\": \"myNode\", \"next\": \"terminate\" }, \"terminate\": { \"type\": \"terminate\" } } } # \"#)?; # Ok::<_, serde_json::Error>(())",
```
mxgrey, line 54: I'm curious: Is there an IDE or other tool that renders this description string, and if so what does this string look like when rendered? As raw text I find it difficult to interpret.

diagram.schema.json
```diff
@@ -0,0 +1,240 @@
...
+          "type": "object",
+          "required": [
+            "next",
+            "nodeId",
```
mxgrey, line 58: I'm a bit reluctant to call this a `nodeId` because
1. Users may think this is supposed to be a UUID rather than an idiomatic name
2. Users may assume every node needs a unique `nodeId`
3. This ID refers to a node **_builder_**, not a node
4. Users may get confused between the purpose of this field and the _name_ of the node (the node's key in the `ops` dictionary)

I would consider using another name for this like `provider` or `builder`.

diagram.schema.json
```diff
@@ -0,0 +1,240 @@
...
+              "enum": [
+                "forkResult"
+              ]
+            }
+          }
+        },
+        {
+          "description": "If the request is a list-like or map-like object, splits it into multiple responses.\n\n# Examples ``` # bevy_impulse::Diagram::from_json_str(r#\" { \"ops\": { \"start\": { \"type\": \"start\", \"next\": \"split\" }, \"split\": { \"type\": \"split\", \"index\": [\"terminate\"] }, \"terminate\": { \"type\": \"terminate\" } } } # \"#)?; # Ok::<_, serde_json::Error>(()) ```",
+          "type": "object",
+          "oneOf": [
```
mxgrey, line 149: We don't need to force users to choose between keys or sequences. I think we can make the split operation look like:

```
{
    "type": "split",
    "keys": {
        # split key -> name of connected node 
        "a": "node_connected_to_a",
        "b": "node_connected_to_b"
    },
    "sequential": [
        "first_node",
        "second_node",
    ]
}
```

The API for split supports both of these kinds of connections simultaneously. `sequential` connections will connect, in sequence, to any split outputs that are not already explicitly connected to by key.

diagram.schema.json
```diff
@@ -0,0 +1,240 @@
...
+            "type": {
+              "type": "string",
+              "enum": [
+                "transform"
+              ]
+            }
+          }
+        },
+        {
+          "description": "Drops the request, equivalent to a no-op.",
```
mxgrey, line 223: I'm a little bit torn on whether `next` should be strictly required field or whether we should interpret the absence of a `next` as a disposal.

In the Rust API we issue a compiler warning if the user lets an output drop without explicitly saying they want to dispose it because it's easy for users to overlook a dropped value that they didn't mean to drop. But I don't think the json schema is where we should try to sniff out those issues. I think it should primarily be the responsibility of the GUI which we will eventually provide for generating these files.

With that in mind, I think the role of the schema should be to help the GUI communicate the user's intention to the executor. What if we change the schema to look like this:

```json
{
    "type": "node",
    "provider": "mul",
    "dispose_output": true,
}
```

So `next` is not strictly required by the schema, and `"dispose_output": true` can be used to indicate that the user has explicitly asked for the output to be disposed (we can give it a default value of `false`). If `next` is missing **and** `dispose_output` is `false` then the GUI and executor can both issue a warning at runtime.

mxgrey (changes requested):
This is looking very impressive! I'm leaving some initial feedback on the JSON schema. I'll keep reviewing this PR tomorrow to provide feedback on the implementation.

I'll go on record as saying that I have a strong dislike of JSON's `camelCase` convention in general. That's not a critique of your work here since I know `camelCase` is the most widely accepted convention for naming fields in the JavaScript/JSON ecosystem. I'm only voicing that I'd personally prefer if we use `snake_case` since I find it much more readable. We also have a precedent set in [`rmf_api_msgs` schemas](https://github.com/open-rmf/rmf_api_msgs/blob/0141ab64e03bf08aba79be76b31ea02a55a3b296/rmf_api_msgs/schemas/robot_commission_request.json#L22) but of course I'm the one that set that precedent so that point should be taken with a grain of salt.

I won't make a big fuss about the `camelCase` if you feel strongly that we should use it, but if you're open to using `snake_case` I'd appreciate the consistency and (subjective) readability.

diagram.schema.json
koonpeng: In vscode it will render the the newlines but it doesn't seem to render the markdown, schemars also doesn't seem to generate the formatting properly. I guess we should remove the examples from the schema to make it play nicer.

mxgrey (changes requested):
These capabilities are looking fantastic, and the implementation is generally very good.

I've left a bunch of inline comments for things that I think could use some tweaking. For larger changes I've opened issues #43 and #44 as follow up items rather than blocking this PR.

diagram.schema.json
koonpeng: ca569172a3e2b1214625b53046be7d95621803ef

mxgrey: Nice, we'll just need to update the [`NodeNotFound`](https://github.com/open-rmf/bevy_impulse/commit/ca569172a3e2b1214625b53046be7d95621803ef#diff-e7190dc1175e279538fa4f9d17a6cbb46119d50677ed0eeea978f26ca17a119bR402) variant to `BuilderNotFound`.

diagram.schema.json
```diff
@@ -0,0 +1,240 @@
...
+            "type"
+          ],
+          "properties": {
+            "next": {
+              "type": "string"
+            },
+            "type": {
+              "type": "string",
+              "enum": [
+                "start"
```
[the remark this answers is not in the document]
koonpeng: I think if we make next optional with automatic implied `dispose` if not specified, then we can avoid reserving a name for `terminate` as well.

src/diagram/generate_schema.rs
```diff
@@ -0,0 +1,35 @@
+use bevy_impulse::Diagram;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let schema = schemars::schema_for!(Diagram);
```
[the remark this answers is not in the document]
koonpeng: There is a test to check if they go out of sync. I'm not too sure about adding a `build.rs` as technically you should not create files outside the build directory to avoid messing up the build cache.

src/diagram/join.rs
```diff
@@ -0,0 +1,106 @@
...
+{
+    debug!("join outputs");
+
+    let first_type = outputs[0].type_id;
+
+    let outputs = outputs
+        .into_iter()
+        .map(|o| {
+            if o.type_id != first_type {
+                Err(DiagramError::TypeMismatch)
```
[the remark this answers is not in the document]
koonpeng: I would like to avoid automatic serialization if possible.

> Otherwise users will only be able to join outputs whose types perfectly match, but I expect in most cases they'll want to join different output types to create one data structure that combines both types.

In that case, they can use the `transform` op to convert it to a compatible type, it does a serialization and a deserialization. But right now the lack of support for serializing joined outputs is stopping the ability to compose something out of multiple outputs.

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,807 @@
...
+            scope: output.scope(),
+            target: output.id(),
+            type_id: TypeId::of::<T>(),
+        }
+    }
+}
+
+#[derive(Clone, Serialize)]
+pub(super) struct NodeMetadata {
+    pub(super) id: &'static str,
```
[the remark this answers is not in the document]
koonpeng: I think the idea is exactly to disallow generating node builders at run time. The builder id should be consistent and never change. The way it is designed now, it is not possible to generate a builder at run time, builders are callbacks that return a node, to generate a builder at run time would require compiling code.

src/diagram/join.rs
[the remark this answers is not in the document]
mxgrey: I think we can move this conversation to #44, but being able to join multiple different data types is for sure going to be a crucial feature.

diagram.schema.json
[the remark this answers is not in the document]
mxgrey: I see what you mean but I'd much prefer if `terminate` was explicit somehow due to how significant it is to the workflow.

I'd rather have `"terminate": true` than nothing at all, but I'm a bit concerned about how people would interpret this:

```json
{
    "next": "something",
    "terminate": true
}
```

I suppose we could produce an error for that.

One option could be:

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "next")]
enum Next {
    Connect { target: OperationId },
    Terminate,
    Dispose,
}
```

If we use that with `#[serde(flatten)]` on the `next` field then we should get a pattern like this:

```json
{
    "foo": {
        "type": "node",
        "builder": "foo",
        "next": "connect",
        "target": "bar"
    },
    "bar": {
        "type": "node",
        "builder": "bar",
        "next": "terminate"
    }
}
```

However this would add some complexity for things like `fork_clone` and `fork_result`, e.g.:

```json
{
    "foo": {
        "type": "fork_clone",
        "next": [
            { "next": "connect", "target": "fork_result" },
            { "next": "connect", "target": "bar" }
        ]
    },
    "fork_result": {
        "type": "fork_result",
        "ok": {
            "next": "connect",
            "target": "foo"
        },
        "err": {
            "next": "terminate"
        }
    }
}
```

We could forbid `fork_clone` from special connections like terminate or disposing, but that might not work well if some other special operation comes along, such as `stream_out` which I mention in #43.

What if we have an `op:` string prefix for special operations, so the previous example could look like:

```json
{
    "foo": {
        "type": "fork_clone",
        "next": ["fork_result", "bar"]
    },
    "fork_result": {
        "type": "fork_result",
        "ok": "foo",
        "err": "op:terminate"
    }
}
```

Every string beginning with `op:` is considered reserved for builtin operations of `bevy_impulse` so we can add `op:stream_out` and others in the future without any risk of breaking backwards compatibility.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
mxgrey: We absolutely should not get in the way of generating node builders at runtime. That's an artificial limitation with no benefit that we'd be imposing due to a lack of imagination on our part.

There is no reason that a user shouldn't be allowed to generate a node builder at runtime, especially if the builder doesn't need to register a continuous service. There could be node builders that use closures which capture runtime data which gets streamed in while the application is running. Then the remote client could ask for that specific node builder to be used in a diagram.

diagram.schema.json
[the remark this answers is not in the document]
koonpeng: I would prefer that we avoid reserving special names because that messes up schema validation. A `oneOf` with any string and some reserved enums would always pass validation even if the op does not exist. Json schema does have a `pattern` that specify a regex that the string must match, but I still prefer to avoid that if possible.

But if we were to reserve names, I think it is better to avoid special characters like `:`, they could cause problems in the future. `_` and `-` would be the safest but less ergonomic.

diagram.schema.json
koonpeng: 01d3bb61e3a429374caf2ca9ebde931932c14c71

src/diagram.rs
```diff
@@ -0,0 +1,844 @@
...
+    NotUnzippable,
+    CannotForkResult,
+    NotSplittable,
+    CannotTransform(TransformError),
+    BadInterconnectChain,
+    JsonError(serde_json::Error),
+    ConnectionError(Box<dyn Error>),
+}
+
+impl Display for DiagramError {
```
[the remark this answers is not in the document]
koonpeng: a74223c2379ac1feac595ed3deb7f729e9f89295

src/diagram/serialization.rs
```diff
@@ -0,0 +1,261 @@
...
+
+use super::NodeRegistry;
+
+#[derive(Debug)]
+pub enum SerializationError {
+    NotSupported,
+    JsonError(serde_json::Error),
+}
+
+impl Display for SerializationError {
```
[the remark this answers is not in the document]
koonpeng: 35a1bb741e81f972663c7445d71dccc31628acbc

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,757 @@
...
+    >
+{
+    pub fn new(registry: &'a mut NodeRegistry) -> Self {
+        Self {
+            registry,
+            _unused: Default::default(),
+        }
+    }
+
+    pub fn register_node<Config, Request, Response, Streams: StreamPack>(
```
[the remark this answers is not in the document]
koonpeng: f529f5cd1392402115cc319c681a8acb3f5d8c15

examples/diagram/calculator/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+        builder.create_map_block(move |req: f64| req - config)
+    });
+    registry.register_node("mul", "Multiply", |builder, config: f64| {
+        builder.create_map_block(move |req: f64| req * config)
+    });
+    registry.register_node("div", "Divide", |builder, config: f64| {
+        builder.create_map_block(move |req: f64| req / config)
+    });
+
+    let mut args = std::env::args();
```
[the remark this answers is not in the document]
koonpeng: 3cbbd137c922cbe4616ab108a3f481723ffdd4cf

examples/diagram/calculator/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    args.next();
+    let path = args.next().unwrap();
+    let request_str = args.next().unwrap();
+
+    let mut app = bevy_app::App::new();
+    app.add_systems(Update, flush_impulses());
```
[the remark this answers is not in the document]
koonpeng: 3cbbd137c922cbe4616ab108a3f481723ffdd4cf

src/diagram.rs
```diff
@@ -0,0 +1,865 @@
...
+    ///         }
+    ///     }
+    /// }
+    /// "#;
+    ///
+    /// let diagram = Diagram::from_json_str(json_str)?;
+    /// let workflow = diagram.spawn_io_workflow(&mut app, &registry)?;
+    /// # Ok::<_, DiagramError>(())
+    /// ```
+    pub fn spawn_workflow<Streams>(
```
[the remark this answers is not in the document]
koonpeng: Should we remove `spawn_workflow` instead, and only have `spawn_io_workflow` to keep consistency? Also if/when we support streams, then we will not have naming conflicts.

diagram.schema.json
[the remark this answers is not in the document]
mxgrey: I believe I've come up with a schema that can satisfy all concerns:

```rust
#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Hash, PartialEq, Eq, PartialOrd, Ord)]
#[serde(untagged, rename_all = "snake_case")]
pub enum OperationId {
    Target(String),
    Builtin { builtin: BuiltinOperation },
}

/// These operations are built into `bevy_impulse` and therefore do not require
/// the `ops` section to instantiate it.
#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Hash, PartialEq, Eq, PartialOrd, Ord)]
#[serde(rename_all = "snake_case")]
pub enum BuiltinOperation {
    /// Use the output to terminate the workflow. This will be the return value
    /// of the workflow.
    Terminate,
    /// Dispose of the output.
    Dispose,
}
```

In terms of what the json would look like, we'd get this:

```json
{
    "foo": {
        "type": "fork_clone",
        "next": ["fork_result", "bar"]
    },
    "fork_result": {
        "type": "fork_result",
        "ok": "foo",
        "err": { "builtin": "terminate" }
    },
    "bar": {
        "type": "node",
        "builder": "bar",
        "next": { "builtin": "dispose" }
    }
}
```

Regular strings would just be operation IDs within the `ops` section, whereas `{ "builtin": "x" }` would be for unambiguous builtin operations which are explicitly enumerated and therefore can be validated.

src/diagram.rs
```diff
@@ -0,0 +1,865 @@
...
+        &self,
+        app: &mut App,
```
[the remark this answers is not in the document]
koonpeng: 6d8340f0c372e68fe364062b28221f1ecae5717e

src/diagram.rs
```diff
@@ -0,0 +1,865 @@
...
+    /// ```
+    Transform(TransformOp),
+
+    /// Drops the request, equivalent to a no-op.
+    Dispose,
+}
+
+type ScopeStart = serde_json::Value;
+type ScopeTerminate = serde_json::Value;
+type DynScope<Streams> = Scope<ScopeStart, ScopeTerminate, Streams>;
```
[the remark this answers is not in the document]
koonpeng: 6d8340f0c372e68fe364062b28221f1ecae5717e

src/diagram.rs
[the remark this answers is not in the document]
mxgrey: I'm hoping that the absence of stream support will be temporary and that we won't do a release until we get it in. I don't think it will take long to implement, I'm just punting on it for right now since this PR is already very beefy. Once we have stream support, all workflows generated by diagrams should have `DiagramStream` as its stream type, so there would no longer be a need for `spawn_io_workflow`.

That being said, if you're concerned that we could suddenly have a spike in users of the `main` branch in between the time we merge this PR and merge the next one then we could go ahead and keep `spawn_io_workflow` for now and get rid of `spawn_workflow` until streams are supported.

src/diagram.rs
```diff
@@ -0,0 +1,844 @@
...
+    ///             "type": "terminate"
+    ///         }
+    ///     }
+    /// }
+    /// # "#)?;
+    /// # Ok::<_, serde_json::Error>(())
+    /// ```
+    Transform(TransformOp),
+
+    /// Drops the request, equivalent to a no-op.
```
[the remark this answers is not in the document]
koonpeng: cf121eadc09eb283845fd7ace76aa2c5d5bbe600

src/diagram.rs
```diff
@@ -0,0 +1,844 @@
...
+    ///             "type": "terminate"
+    ///         }
+    ///     }
+    /// }
+    /// # "#)?;
+    /// # Ok::<_, serde_json::Error>(())
+    /// ```
+    Join(JoinOp),
+
+    /// If the request is serializable, transforms it by running it through a [CEL](https://cel.dev/) program.
```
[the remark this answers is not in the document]
koonpeng: cf121eadc09eb283845fd7ace76aa2c5d5bbe600

src/diagram.rs
```diff
@@ -0,0 +1,844 @@
...
+    ///             "type": "terminate"
+    ///         }
+    ///     }
+    /// }
+    /// # "#)?;
+    /// # Ok::<_, serde_json::Error>(())
+    /// ```
+    Split(SplitOp),
+
+    /// Waits for an item to be emitted from each of the inputs, then combined the
```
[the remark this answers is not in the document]
koonpeng: cf121eadc09eb283845fd7ace76aa2c5d5bbe600

src/diagram.rs
```diff
@@ -0,0 +1,844 @@
...
+    ///         "terminate": {
+    ///             "type": "terminate"
+    ///         }
+    ///     }
+    /// }
+    /// # "#)?;
+    /// # Ok::<_, serde_json::Error>(())
+    ForkResult(ForkResultOp),
+
+    /// If the request is a list-like or map-like object, splits it into multiple responses.
```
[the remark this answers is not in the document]
koonpeng: cf121eadc09eb283845fd7ace76aa2c5d5bbe600

diagram.schema.json
mxgrey: If we follow through on [this recommendation](https://github.com/open-rmf/bevy_impulse/pull/27#discussion_r1901449661) then we can keep `next` as required and any disposed outputs must be set to `{ "builtin": "dispose" }`.

src/diagram.rs
```diff
@@ -0,0 +1,807 @@
...
+
+    #[error("the number of unzip slots in response does not match the number of inputs")]
+    NotUnzippable,
+
+    #[error(
+        "node must be registered with \"with_fork_result()\" to be able to perform fork result"
+    )]
+    CannotForkResult,
+
+    #[error("response cannot be splitted")]
```
mxgrey, line 431:
```suggestion
    #[error("response cannot be split")]
```

split is one of those weird English words where it is its own past tense

src/diagram.rs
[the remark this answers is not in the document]
koonpeng: > Once we have stream support, all workflows generated by diagrams should have DiagramStream as its stream type, so there would no longer be a need for spawn_io_workflow.

I think there might be cases where there will be other streams. I see streams as something that does not necessarily need to come from the request to a workflow, an app that uses diagram could have some default/universal stream, in those cases, streams would be opaque as we never have to do any serialization.

diagram.schema.json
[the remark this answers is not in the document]
koonpeng: I think this looks good :+1:

src/diagram/fork_result.rs
```diff
@@ -0,0 +1,144 @@
...
+    E: Send + Sync + 'static,
+{
+    const SUPPORTED: bool = true;
+
+    fn dyn_fork_result(
+        builder: &mut Builder,
+        output: DynOutput,
+    ) -> Result<(DynOutput, DynOutput), DiagramError> {
+        debug!("fork result: {:?}", output);
+        assert_eq!(output.type_id, TypeId::of::<Result<T, E>>());
```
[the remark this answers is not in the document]
koonpeng: c23ffe414d9d60f4f2063ec1fbb39a4ee5c7b906

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,807 @@
...
+
+/// A type erased [`crate::Output`]
+pub struct DynOutput {
+    scope: Entity,
+    target: Entity,
+    pub(super) type_id: TypeId,
+}
+
+impl DynOutput {
+    pub(super) fn into_output<T>(self) -> Output<T>
```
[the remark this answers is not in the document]
koonpeng: c23ffe414d9d60f4f2063ec1fbb39a4ee5c7b906

src/diagram/join.rs
```diff
@@ -0,0 +1,106 @@
...
+    register_serialize::<Vec<T>, Serializer>(registry);
+}
+
+fn join_impl<T>(builder: &mut Builder, outputs: Vec<DynOutput>) -> Result<DynOutput, DiagramError>
+where
+    T: Send + Sync + 'static,
+{
+    debug!("join outputs");
+
+    let first_type = outputs[0].type_id;
```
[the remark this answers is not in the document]
koonpeng: c23ffe414d9d60f4f2063ec1fbb39a4ee5c7b906

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,807 @@
...
+#[derive(Copy, Clone)]
+pub struct DynInputSlot {
+    scope: Entity,
+    source: Entity,
+    pub(super) type_id: TypeId,
+}
+
+impl DynInputSlot {
+    pub(super) fn into_input<T>(self) -> InputSlot<T> {
+        InputSlot::<T>::new(self.scope, self.source)
```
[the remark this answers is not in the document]
koonpeng: c23ffe414d9d60f4f2063ec1fbb39a4ee5c7b906

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,807 @@
...
+    }
+}
+
+impl Debug for DynInputSlot {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
```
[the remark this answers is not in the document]
koonpeng: c23ffe414d9d60f4f2063ec1fbb39a4ee5c7b906

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,807 @@
...
+    where
+        T: Send + Sync + 'static,
+    {
+        Output::<T>::new(self.scope, self.target)
+    }
+}
+
+impl Debug for DynOutput {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
```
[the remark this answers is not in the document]
koonpeng: c23ffe414d9d60f4f2063ec1fbb39a4ee5c7b906

src/diagram.rs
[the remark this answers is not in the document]
mxgrey: I do have some thoughts on how it may be possible to support other concrete stream types, but it would be related to also supporting concrete request and response types, and it would require some reimplementation of the `Scope` operation. I would save that for a far future PR.

src/diagram/join.rs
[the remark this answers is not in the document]
mxgrey: We shouldn't allow an empty join, we should treat it as an error. An empty join can never be triggered which makes everything connected to it effectively dead. If a user has an empty join in their diagram then they have made some kind of mistake and we should give them a specific error to let them know.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
koonpeng: > I could write an example of a bevy system that could take runtime data from an event and move it into a closure to produce and register a new node builder if you'd like.

I guess it is up for debate whether this would be considered a run time builder or still compile time. Technically there is only 1 closure (assuming no generics at play) that captures runtime values. I guess that there could be some use cases to do this over using configs.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
koonpeng: f529f5c

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,807 @@
...
+        Self {
+            registry,
+            _unused: Default::default(),
+        }
+    }
+
+    pub fn register_node<Config, Request, Response, Streams: StreamPack>(
+        &mut self,
+        id: &'static str,
+        name: &'static str,
```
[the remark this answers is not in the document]
koonpeng: 7a825658dfc12a5e58380e1b54e68ddf7f4e36ee

src/diagram.rs
koonpeng: bf63e5b20290473dbc34dc95a1e0057422d8499d

src/diagram/node_registry.rs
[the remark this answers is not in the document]
mxgrey: It's true that ultimately **_something_** needs to be compiled ahead of time, but there are many ways you could make that **_something_** very dynamic. Here's one example:

```rust
fn create_new_node_builder(
    builder_name: String,
    service: Service<serde_json::Value, serde_json::Value>,
    registry: &mut NodeRegistry,
) {
    registry.register_node_builder(
        builder_name.clone(),
        builder_name.clone(),
        move |builder, config: ()| {
            builder.create_node(service.clone())
        },
    );
}
```

In this case `service` may itself be something generated at runtime, e.g. a workflow which itself gets built from a diagram description. Users might want to register their custom tasks as node builders, and those custom tasks may be stored in databases that need to be loaded at runtime.

src/diagram/generate_schema.rs
```diff
@@ -0,0 +1,35 @@
...
+        #[test]
+        fn check_schema_changes() -> Result<(), String> {
+            let cur_schema_json = std::fs::read("diagram.schema.json").unwrap();
+            let schema = schemars::schema_for!(Diagram);
+            let new_schema_json = serde_json::to_vec_pretty(&schema).unwrap();
+
+            if cur_schema_json.len() != new_schema_json.len()
+                || zip(cur_schema_json, new_schema_json).any(|(a, b)| a != b)
+            {
+                return Err(String::from("There are changes in the json schema, please run `cargo run -F=diagram generate_schema` to regenerate it"));
```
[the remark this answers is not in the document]
mxgrey: We'll go forward with the test approach for now and see how it feels. If we run into friction down the road we can revisit the idea of fully automating the generating.

src/diagram/testing.rs
```diff
@@ -0,0 +1,119 @@
...
+        "multiply3",
+        "multiply3",
+        |builder: &mut Builder, _config: ()| builder.create_map_block(multiply3),
+    );
+    registry
+        .registration_builder()
+        .with_response_cloneable()
+        .register_node(
+            "multiply3_cloneable",
+            "multiply3_cloneable",
```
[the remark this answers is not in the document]
koonpeng: c5ced4519a688c399e9529a91b9ac75a6a895e88

src/diagram/transform.rs
```diff
@@ -0,0 +1,230 @@
...
+    } else {
+        let serialize = registry
+            .serialize_impls
+            .get(&output.type_id)
+            .ok_or(DiagramError::NotSerializable)?;
+        serialize(builder, output)
+    };
+
+    let program = Program::compile(&transform_op.cel).map_err(|err| TransformError::Parse(err))?;
+    let transform_node = builder.create_map_block(
```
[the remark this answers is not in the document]
koonpeng: > terminating: CEL programs cannot loop forever;

I saw on the docs that CEL is designed to be terminating https://github.com/google/cel-spec/blob/master/doc/langdef.md#list-of-standard-definitions. It does reserve words like "loop" and "while" but it seems to be just to avoid conflicts when embedding CEL.

I think in most cases we can assume CEL programs to be short and simple, for complex transformations that can take a long time, the better approach would be to have a node that does the transformation, using native rust code instead of CEL.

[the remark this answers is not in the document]
mxgrey: Sure, we can go with this as-is, and if we ever do find a use case for long-running CEL operations we can always add an async option in a backwards-compatible way.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
mxgrey: It says `name` is used for display purposes, but where does it actually get used?

If it's meant to be displayed in workflow visualizations in the future then I'd suggest the following:
1. Remove this argument from the `register_node_builder` API
2. Support a `"label": "simple name"` in the `"type": "node"` operation

I think it will make more sense for human-readable node names to be decided inside the diagram file than by the `register_node_builder` call. For example if you have multiple instances generated by the same node builder, you may want them to have different human-readable names, which the current API would not support.

src/diagram/node_registry.rs
```diff
@@ -0,0 +1,807 @@
...
+    ///     .with_opaque_response()
+    ///     .register_node("echo", "echo", |builder, _config: ()| {
+    ///         builder.create_map_block(|msg: NonSerializable| msg)
+    ///     });
+    /// ```
+    ///
+    /// Note that nodes registered this way cannot be connected to "Start" or "Terminate".
+    pub fn registration_builder(
+        &mut self,
+    ) -> RegistrationBuilder<
```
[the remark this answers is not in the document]
koonpeng: Yeah the reason for this is to support opt-out.

I think there is lots of ways to design this part of the code, I'm inclined to say that we need a long discussion to find the best approach. The "problem" is that refactoring this will likely result in the change of the public api.

Some ideas I had:

* Having some operations like fork clone opt-out, similar to serialization. For now it's just fork clone and fork result I'm considering but with the above design, I find it less ergonomic that this change will require refactoring instead of just flipping a switch.
* A way to reduce the amount of generic params is to group ops together, but that doesn't really reduce the complexity, we would get params like `ForkCloneAndUnzip<NotSupported, DefaultImpl>`.
* Perhaps the most rustic approach would be to use derive macros. Technically serialization is indirectly "opt-in" because you need to derive `JsonSchema` and `Serialize`, so it makes sense to derive an `OpaqueMessage`. But should we extend it to other operations? Something like

```rust
#[derive(DiagramMessage)]
#[diagram_message(serialize = false)
struct Foo {}
```

The downside is that every request and response would need to `derive`, but even with the current approach they most likely need to derive `JsonSchema` and `Serialize` already anyway. I think a huge advantage is that it reduces the occurrence of incomplete flagging, e.g. a `Result<_, _>` not being flagged with `with_fork_result`. It also allows to automatically impl some operations like unzip and fork result https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2be9b765e3346bc29bec6397f224f616.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
koonpeng: [7a82565](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

koonpeng:
The debug print outs are actually intended, I think they will be useful for diagnostics. They won't appear unless the user

1. set ups a tracing emitter
2. turns on debug level tracing

bevy uses the same framework and prints some logs as well

e.g. `RUST_LOG=debug cargo test -F=diagram diagram -- --show-output`
```
...
2025-01-06T04:44:56.577813Z DEBUG bevy_ecs::world::entity_ref: Despawning entity 10v2
2025-01-06T04:44:56.577870Z DEBUG bevy_ecs::world::entity_ref: Despawning entity 26v0
2025-01-06T04:44:56.577918Z DEBUG bevy_ecs::world::entity_ref: Despawning entity 9v1
2025-01-06T04:44:56.577959Z DEBUG bevy_ecs::world::entity_ref: Despawning entity 29v0
2025-01-06T04:44:56.578001Z DEBUG bevy_ecs::world::entity_ref: Despawning entity 23v0
2025-01-06T04:44:56.578036Z DEBUG bevy_ecs::world::entity_ref: Despawning entity 32v0
```

src/diagram.rs
```diff
@@ -0,0 +1,844 @@
...
+// ----------
+
+use std::{collections::HashMap, error::Error, fmt::Display, io::Read};
+
+use crate::{Builder, Scope, Service, SpawnWorkflowExt, SplitConnectionError, StreamPack};
+use bevy_app::App;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+pub type NodeId = String;
```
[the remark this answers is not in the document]
koonpeng: ca56917

src/diagram/node_registry.rs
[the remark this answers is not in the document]
mxgrey: This whole problem would be extremely easy with the `min_specialization` language feature, but there's no telling how much longer before that features is available in stable Rust, so sadly yeah we're going to need to figure out the least objectionable set of trade-offs to solve this problem.

Using `derive` is worth considering, but derive macros are unable to infer traits of the structs and enums that they are used on (nor on the fields of those structs and enums), because derive macros run before trait resolution is performed. That puts some significant limits on how much we can automate with macros. And then accounting for how it will interfere with users' ability to send and receive data types that are independent of `bevy_impulse`, I find it hard to justify that approach.

> group ops together ... `ForkCloneAndUnzip<NotSupported, DefaultImpl>`

I don't think this will meaningfully reduce the overall complexity of the generics, it would just shift it deeper and probably make the overall setup harder to understand.

> Having some operations like fork clone opt-out, similar to serialization. For now it's just fork clone and fork result I'm considering but with the above design, I find it less ergonomic that this change will require refactoring instead of just flipping a switch.

In the setup that I'm proposing, opt-out operations would go into `OpacityNodeBuilder`. If we anticipate we'll want to add more opt-out operations in the future then we can consider changing the term "opacity" to something else, e.g.:

```rust
pub struct DefaultOps<Deserializer, Serializer, Cloneable> {
    ...
}

impl<S, C> DefaultOps<DefaultDeserializer, S, C> {
    pub fn no_request_deserialize(self) -> DefaultOps<NotSupported, S, C> { ... }
}

impl NodeRegistry {
    pub fn opt_out(&mut self) -> DefaultOps<DefaultDeserializer, DefaultSerializer, DefaultClone> {
        ...
    }

    pub fn register_node_builder< ... >( ... ) -> RegistrationBuilder<Request, Response, Streams>
    where
        ...
    {
        self.opt_out().register_node_builder()
    }
```

With that approach, we would only need to add new generics for `DefaultOps`, which I think would happen very rarely since the vast majority of operations are only applicable to specific data types. If we make `fork_clone` a default operation then we're still capping out our generic arguments at 3 per struct even if we keep adding non-default operations in the future.

src/diagram/join.rs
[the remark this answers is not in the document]
koonpeng: e3e1fb7801705010624e60823a35a049cad496a1

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,501 @@
...
+    inputs: &HashMap<&OperationId, DynInputSlot>,
+    target: &'a Vertex,
+) -> Result<(), DiagramError> {
+    debug!("connecting [{}]", target.op_id);
+    match target.op {
+        // join needs all incoming edges to be connected at once so it is done at the vertex level
+        // instead of per edge level.
+        DiagramOperation::Join(_) => {
+            if target.in_edges.is_empty() {
+                if target.in_edges.is_empty() {
```
mxgrey, line 235: This seems redundant with the previous check

src/diagram/join.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    let outputs = outputs
+        .into_iter()
+        .map(|o| {
+            // joins is only supported for outputs of the same type. This is because joins of
+            // different types produces a tuple and we cannot output a tuple as we don't
+            // know the number and order of join inputs at compile time.
+            // A workaround is to serialize them all the `serde_json::Value` or convert them to `Box<dyn Any>`.
+            // But the problem with `Box<dyn Any>` is that we can't convert it back to the original type,
+            // so nodes need to take a request of `JoinOutput<Box<dyn Any>>`.
```
mxgrey, line 59: This might be too cumbersome for users in many cases.

What if we do:

```
{
    "type": "join",
    "next": "foo",
    "serialize": true
}
```

so users can decide within the diagram to send a serialized version of the data to the next node. I think it'll be much less troublesome for users to create nodes that support deserializing than nodes that can only connect to `JoinOutput<T>`.

src/diagram/workflow_builder.rs
koonpeng: 08631b4c15a8dded54452e4783d0bfb80dea83f0 oops, copy paste error

diagram.schema.json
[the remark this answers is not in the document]
koonpeng: 6a20ef52d1990be529b1d2149316a91c361fab04

src/diagram.rs
```diff
@@ -0,0 +1,669 @@
...
+type DiagramStart = serde_json::Value;
+type DiagramTerminate = serde_json::Value;
+type DiagramScope<Streams = ()> = Scope<DiagramStart, DiagramTerminate, Streams>;
+
+#[derive(JsonSchema, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct Diagram {
+    /// Version of the diagram, should always be `1`.
+    #[schemars(range(min = 1, max = 1))]
+    version: u32,
```
mxgrey, line 276: I like the idea of a built-in version number, but I think we should use a [`SemVer`](https://github.com/open-rmf/rmf_site/blob/main/rmf_site_format/src/semver.rs#L38) so we can distinguish between forward-compatible and incompatible format changes.

I wrote that `SemVer` class several years ago so it may be worth checking if there's a small crate out in the ecosystem that implements the same concept.

mxgrey: This crate may be worth considering: 

https://docs.rs/serde-semver/latest/serde_semver/

The only thing I don't like is it doesn't seem to let us inspect the actual version number that was present in the serialized data.

koonpeng: 9a231e726654ac73f9e9dfb7c5b50d89fee5f01a

Personally I don't think semver is needed, but it wouldn't hurt also so I changed it.

src/diagram/join.rs
koonpeng: 41a123b9e1d340ebfea021b82013237ce8ab095f

mxgrey: What would you think of the `serialize` option being true by default? I suspect that's what users will want more often than not since the non-serialized version requires them to expect a specific size of `SmallVec`.

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,517 @@
...
+) -> Result<(), DiagramError> {
+    debug!("connecting [{}]", target.op_id);
+    match target.op {
+        // join needs all incoming edges to be connected at once so it is done at the vertex level
+        // instead of per edge level.
+        DiagramOperation::Join(join_op) => {
+            if target.in_edges.is_empty() {
+                return Err(DiagramError::EmptyJoin);
+            }
+            let outputs: Vec<DynOutput> = target
```
mxgrey, line 255: I just thought of a tricky issue with join...

In many cases the order of the inputs going into the join will matter. This is especially the case when the node after the join is expecting a heterogeneous mix. For example if the node after the join expects an input of `(Pose, Velocity)` then it will fail to deserialize if we pass along a `Value::Array([velocity, pose])`.

I think we need to introduce a new field to the `join` operation:

```json
{
    "type": "join",
    "inputs": ["foo", "bar"],
    "next": "baz"
}
```

We'll use the `inputs` field to determine the order that the inputs are arranged in the output. If there's ever a mismatch between the values in `inputs` and the `next` value of one of those inputs, then that should be treated as an error.

This will also be important in the future when we add support for buffers. Buffers are the real inputs for a `join` operation (`join` can take in a bunch of `Output`s, but it actually changes them into buffers before joining them). Buffers won't be able to specify a `next` at all, so the `join` operation will need to be told which buffers to pull from, and we can do that with the `inputs` field.

src/diagram/join.rs
koonpeng: I think it mostly depends on the type of app and the type of user. The end user (the user of the app that uses bevy_impulse diagram) would probably not care about perf, defaulting to serialize would make writing diagrams easier. The users of bevy_impulse (the devs of the app) might not want to end users to serialize, but perhaps defaulting to serialize is one way to push them to have their own "join" node specialized for the data they are handling.

koonpeng: 922db287189996ee529beb13bf38c4f901b01141

src/diagram/workflow_builder.rs
koonpeng: I'm not sure I understand this, I assume the items in the `inputs` array are referring to another operation id? But I'm not sure how we can use that to make the output of join stable.

mxgrey: Correct, the items in `inputs` would be the IDs of other operations (which would also include the names of buffers). The output of join would be stable because we would always arrange the values in the array in the order of the names specified by `inputs`. So if we have `"inputs": ["pose", "velocity"]` where `pose` is outputting a `Pose` and `velocity` is outputting a `Velocity` then we can be confident that the output of `join` will always be a `Value::Array([Pose, Velocity])` which can then be deserialized to a `(Pose, Velocity)`.

Remember that the `join` operation will wait until at least one value of each output is available and then grab the oldest of each to produce the output of the `join`. We just need to make sure that when it grabs the oldest of each, it will arrange them in the array in the order that's expected by the next operation.

koonpeng: 064ec4e84db4e23f9f3984c8f1c19f9f23d75b57

I stablized the order of the outputs passed to `join_vec`, it seems to work for `test_serialize_and_join` but `test_join` is still unstable, could it be due to the extra nodes in `test_join`?

src/diagram/join.rs
```diff
@@ -0,0 +1,314 @@
...
+};
+
+#[derive(Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct JoinOp {
+    pub(super) next: NextOperation,
+
+    /// Controls the order of the resulting join. Each item must be an operation id of one of the
+    /// incoming outputs.
+    pub(super) order: Vec<SourceOperation>,
```
mxgrey, line 21: I understand why you named this `order`, but it will make more sense for it to be named `inputs` or `sources` once we add support for buffers. Buffers won't have a concept of `next`, so this field will be the only place where we will see a connection between a `buffer` and a `join`. At that point it's describing more than just an order; it's describing what its inputs are.

src/diagram/workflow_builder.rs
mxgrey: This one was quite a doozy to figure out. I did a bunch of debug tracing all the way down into the weeds of the `join_vec` implementation trying to find where the outputs could be getting crossed, only to find that it was always working perfectly.

Eventually I tracked the problem down to [this code for the split operation](https://github.com/open-rmf/bevy_impulse/blob/632dfc236f8f57307586cb3684d1815494da534f/src/diagram/workflow_builder.rs#L431-L438). The split outputs are being generated correctly [here](https://github.com/open-rmf/bevy_impulse/blob/632dfc236f8f57307586cb3684d1815494da534f/src/diagram/workflow_builder.rs#L418) as a map from `NextOperation` to the relevant `Output` from the split. However on [this line](https://github.com/open-rmf/bevy_impulse/blob/632dfc236f8f57307586cb3684d1815494da534f/src/diagram/workflow_builder.rs#L435) the `NextOperation` is discarded and replaced with an arbitrary entry from `target.out_edges`.

Changing that block to

```rust
for e in &target.out_edges {
    let out_edge = edges.get_mut(e).unwrap();
    let output = outputs.outputs.remove(out_edge.target).unwrap();
    out_edge.state = EdgeState::Ready { output, origin };
}
```

and making `outputs` mutable fixed the stability issue. Essentially the split operation was randomly pairing each of its outputs to any operation that was connected to it instead of the one that the user specifically asked for.

It's a good thing we had these tests or else this could've easily been missed. I might recommend making a smoke test that does a big split (maybe ~10 values) followed by a big join of all those values, and make sure the final result has every element in the correct location. A really good smoke test would rerun the test numerous times with a randomized set of operation names, since the hashmaps used in the implementation could scramble the orderings based on the name values.

koonpeng: Thanks for helping to figure this out! Yeah, I can see how split causes the ordering to go wrong.

koonpeng: 632dfc236f8f57307586cb3684d1815494da534f...e826741a0de362e1b34c58dea3c4cfa937e50fcb

src/diagram/node_registry.rs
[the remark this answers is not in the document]
koonpeng: To be more specific, it is the node builder's name. Putting a label in the diagram would be more suitable for the operation's name. I am thinking of a scenario where you have a list of node builders available, and you drag one of them to the canvas to add it into a diagram, this is one way that the name would be used.

diagram.schema.json
koonpeng: One reason why I avoided this is because json parsers/interpreters can be finicky. Some of them may allow you to access an array via a key e.g. `arr["1"]`. It does not exactly conflict with dual key-seq support, but it does make things confusing.

```json
{
    "type": "split",
    // an array is actually passed in
    "keys": {
        "1": "node_connected_to_a", // refers to the 2nd item
    },
    "sequential": [
        "first_node", // first item
        "second_node" // third(?) item
    ]
}
```

I wonder in practice how useful dual key-seq would be compared to key + remaining to get the other items. key + remaining does have a downside that you can't grab a subset of the unpicked keys (possible if you connect the output of remaining to another split), but if the input is a `std::HashMap`, the order of the seq outputs would not be stable.

mxgrey: I agree that it would be strange to see situations where people mix both keyed and sequential connections. But since there is no technical reason to prohibit it, and there's no additional difficulty in us supporting it, I would rather leave it up to users to decide if/when simultaneous access to both may be useful for them.

Also I'm thinking we should name the field `keyed` instead of `keys` since I think `keyed` as an adjective aligns better with the adjective `sequential`.

mxgrey: FWIW when it comes to RMF workflows and especially task descriptions, I'm planning on adding an opinionated layer between `bevy_impulse` and the RMF user. That additional layer would add some guard rails around how tasks and RMF workflows are built.

Meanwhile for `bevy_impulse` we should try to keep it as general purpose (unopinionated) as possible so it can be used beyond what we're targeting for RMF.

src/diagram/join.rs
```diff
@@ -0,0 +1,322 @@
...
+            Ok(serialized_output)
+        })
+        .collect::<Result<Vec<_>, DiagramError>>()?;
+
+    // we need to convert the joined output to [`serde_json::Value`] in order for it to be
+    // serializable.
+    let joined_output = outputs.join_vec::<4>(builder).output();
+    let json_output = joined_output
+        .chain(builder)
+        .map_block(|o| serde_json::to_value(o).unwrap())
```
mxgrey, line 73: What are the conditions in which this `.unwrap()` might panic? We should either document a rationale for why a panic is impossible, or else we should use `.cancel_on_err()` here instead of `.unwrap()`.

src/diagram.rs
```diff
@@ -0,0 +1,771 @@
...
+    gen.subschema_for::<String>()
+}
+
+/// deserialize semver and validate that it has a supported version
+fn deserialize_semver<'de, D>(de: D) -> Result<semver::Version, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let s = String::deserialize(de)?;
+    let ver_req = semver::VersionReq::parse(SUPPORTED_DIAGRAM_VERSION).unwrap();
```
mxgrey, line 338: This seems like it should return an error rather than unwrapping. If we can guarantee that the unwrap won't panic then we need to include a comment providing the rationale.

src/diagram/split_serialized.rs
```diff
@@ -0,0 +1,583 @@
...
+                    }
+                }
+            }
+            Value::Object(map) => {
+                let mut next_seq = 0;
+                for (name, value) in map.into_iter() {
+                    let key = MapSplitKey::Specific(name);
+                    match dispatcher.outputs_for(&key) {
+                        Some(outputs) => {
+                            let position = JsonPosition::ObjectField(key.specific().unwrap());
```
mxgrey, line 87: I know I wrote this line, but I had already forgotten what the safety rationale is for being able to use `.unwrap()` here. We should add a line of documentation like

```rust
// SAFETY: This key was initialized as MapSplitKey::Specific earlier in the function and is immutable,
// so this method is guaranteed to return Some
```

src/diagram/split_serialized.rs
```diff
@@ -0,0 +1,583 @@
...
+                            outputs.push((position, value));
+                        }
+                        None => {
+                            // No connection to the specific field name, so let's
+                            // check for a sequential connection.
+                            let seq = MapSplitKey::Sequential(next_seq);
+                            next_seq += 1;
+
+                            let position = JsonPosition::ObjectField(key.specific().unwrap());
```
mxgrey, line 96: We should add the same safety documentation here.

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                        builtin: super::BuiltinSource::Start,
+                    },
+                    target: &diagram.start,
+                    state: EdgeState::Ready {
+                        output: scope.input.into(),
+                        origin: None,
+                    },
+                },
+            );
+            vertices.get_mut(&op_id).unwrap().in_edges.push(0);
```
mxgrey, line 91: We should either provide a rationale for why this unwrap cannot panic or this case should be covered as a `DiagramError`.

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                    .map(|v| v.op_id)
+                    .collect::<Vec<_>>()
+            );
+            return Err(DiagramError::BadInterconnectChain);
+        }
+    }
+
+    // connect terminate
+    for edge_id in terminate_edges {
+        let edge = edges.remove(&edge_id).unwrap();
```
mxgrey, line 238: Either provide a safety rationale or return a `DiagramError`.

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+        // instead of per edge level.
+        DiagramOperation::Join(join_op) => {
+            if target.in_edges.is_empty() {
+                return Err(DiagramError::EmptyJoin);
+            }
+            let mut outputs: HashMap<SourceOperation, DynOutput> = target
+                .in_edges
+                .iter()
+                .map(|e| {
+                    let edge = edges.remove(e).unwrap();
```
mxgrey, line 270: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+            }
+
+            let joined_output = if join_op.no_serialize.unwrap_or(false) {
+                let join_impl = &registry.join_impls[&ordered_outputs[0].type_id];
+                join_impl(builder, ordered_outputs)?
+            } else {
+                serialize_and_join(builder, registry, ordered_outputs)?.into()
+            };
+
+            let out_edge = edges.get_mut(&target.out_edges[0]).unwrap();
```
mxgrey, line 293: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+
+fn connect_edge<'a>(
+    builder: &mut Builder,
+    registry: &NodeRegistry,
+    edges: &mut HashMap<usize, Edge<'a>>,
+    inputs: &HashMap<&OperationId, DynInputSlot>,
+    edge_id: usize,
+    target: &Vertex,
+) -> Result<(), DiagramError> {
+    let edge = edges.remove(&edge_id).unwrap();
```
mxgrey, line 318: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                };
+
+                let reg = registry.get_registration(&origin.builder)?;
+                reg.fork_clone(builder, output, amount)
+            }?;
+            outputs
+                .into_iter()
+                .zip(target.out_edges.iter())
+                .for_each(|(o, e)| {
+                    let out_edge = edges.get_mut(e).unwrap();
```
mxgrey, line 364: Safety rationale or `DiagramError`

Also I recommend using conventional for-loops instead of `for_each` if you don't have a very long chain leading up to it. Conventional for-loops make error handling more ergonomic.

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                reg.unzip(builder, output)
+            }?;
+            if outputs.len() < unzip_op.next.len() {
+                return Err(DiagramError::NotUnzippable);
+            }
+            outputs
+                .into_iter()
+                .zip(target.out_edges.iter())
+                .for_each(|(o, e)| {
+                    let out_edge = edges.get_mut(e).unwrap();
```
mxgrey, line 388: Safety rationale or `DiagramError`

Also consider a conventional for-loop

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                    origin_node
+                } else {
+                    return Err(DiagramError::CannotForkResult);
+                };
+
+                let reg = registry.get_registration(&origin.builder)?;
+                reg.fork_result(builder, output)
+            }?;
+            {
+                let out_edge = edges.get_mut(&target.out_edges[0]).unwrap();
```
mxgrey, line 406: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                out_edge.state = EdgeState::Ready { output: ok, origin };
+            }
+            {
+                let out_edge = edges.get_mut(&target.out_edges[1]).unwrap();
```
mxgrey, line 410: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                };
+
+                let reg = registry.get_registration(&origin.builder)?;
+                reg.split(builder, output, split_op)
+            }?;
+
+            // Because of how we build `out_edges`, if the split op uses the `remaining` slot,
+            // then the last item will always be the remaining edge.
+            let remaining_edge_id = if split_op.remaining.is_some() {
+                Some(target.out_edges.last().unwrap())
```
mxgrey, line 435: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                None
+            };
+            let other_edge_ids = if split_op.remaining.is_some() {
+                &target.out_edges[..(target.out_edges.len() - 1)]
+            } else {
+                &target.out_edges[..]
+            };
+
+            for e in other_edge_ids {
+                let out_edge = edges.get_mut(e).unwrap();
```
mxgrey, line 446: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                let output = outputs.outputs.remove(out_edge.target).unwrap();
```
mxgrey, line 447: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                out_edge.state = EdgeState::Ready { output, origin };
+            }
+            if let Some(remaining_edge_id) = remaining_edge_id {
+                let out_edge = edges.get_mut(remaining_edge_id).unwrap();
```
mxgrey, line 451: Safety rationale or `DiagramError`

src/diagram/workflow_builder.rs
```diff
@@ -0,0 +1,547 @@
...
+                    origin,
+                };
+            }
+        }
+        DiagramOperation::Join(_) => {
+            // join is connected at the vertex level
+        }
+        DiagramOperation::Transform(transform_op) => {
+            let transformed_output = transform_output(builder, registry, output, transform_op)?;
+            let out_edge = edges.get_mut(&target.out_edges[0]).unwrap();
```
mxgrey, line 463: Safety rationale or `DiagramError`

mxgrey (changes requested):
This review is primarily an audit of the use of `.unwrap()` inside the critical path.

It's imperative that we bring the risk of crashing/aborting/panicking as close to zero as possible, or else we undermine the crash-free exception-free selling points of Rust.

There will certainly be cases where we can guarantee that `.unwrap()` is perfectly safe, and in those places we should add clear documentation to explain how we know this.

But in general we should prefer to return `Err` results if we aren't positively 100% certain that the unwrap will not panic.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
koonpeng: > Do we expect that to commonly be different from the id?

That would depends on several factors. Right now `id` is just a string with no constraints, so it is very tempting to make that same as `name`, but I believe there are reasons to consider constraining it. Whether user needs `name` would also depends on how they want to manage ids, if they use a UUID string for ids, then they would almost certainly need `name`. A potential use case for doing something like that is the scenario we discussed in the choice of `&'static str`. If they are storing builders in dbs, or generating builders at run time, then there is a good reason to use something like UUIDs or hashes for ids.

An alternative is to use a `label: HashMap<String, String>` to store extra data, but I think `name` is fundamental enough that it should always be required (if we consider frontend as a first class feature).

src/diagram/node_registry.rs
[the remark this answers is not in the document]
koonpeng: While trying to change this, I realized that it will require a pretty large refactor. There are also some things that I will need more time to implement, particularly `DefaultOps::register_node_builder()` needs to pass the serializers etc to `RegistrationBuilder`, can be easily done by adding a generic param to `RegistrationBuilder`, but then we wouldn't really need `opt_out` anymore in that case, we can do something like

```rust
struct RegistrationBuilder<Opacity, Request, Response, Streams> {
  ...
  // these are now needed because we no longer "stores" the impls in generic params.
  fork_clone_impl: Option<Box<dyn DynForkClone>>,
  ...
}

impl<...> RegistrationBuilder<...> {
  fn opaque_request(...) -> RegistrationBuilder<OpacityImpl<Opaque, Default>, ...> {
      ...
  }
}
```

I'm also still not convinced that we should drop the idea of using derives. Derive could potentially allow us to remove the need to opt out/in many operations like unzip, fork result and split, every node can then automatically register impls for those operations. That leaves only serialization and fork clone that still needs manual control. tbh I'm still not certain if that is possible but I do have some ideas and I think it is worth exploring.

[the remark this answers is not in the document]
mxgrey: > Derive could potentially allow us to remove the need to opt out/in many operations like unzip, fork result and split

We can't apply `derive` to any `Result<T, E>` so that immediately rules out its usefulness for `fork_result`. I think we should shelf the derive idea for this initial implementation. There are too many jagged edges.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
mxgrey: Understood. I think we should use the `NodeBuilderOptions` pattern for the sake of forwards API compatibility. I think having an equivalent `id` and `name` will be common enough that users will appreciate being able to just define one of them.

```rust
#[non_exhaustive]
pub struct NodeBuilderOptions {
    pub id: String,
    pub name: Option<String>,
}
```

Then we can add more fields to `NodeBuilderOptions` if we ever want to add more options for the node builders.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
mxgrey: > needs to pass the serializers etc to RegistrationBuilder

I looked into this more. I see that `SerializeImpl` is used by `SplitImpl` and `UnzipImpl`, so I take it that's why you say that `RegistrationBuilder` needs the serializer. I still think `<Opacity, Request, Response, Streams>` is preferable to the six generic arguments that it currently needs, especially since that number would be prone to grow if we add more operations.

Another option would be to offer variations for `split` and `unzip` that don't require `Opacity`:

```rust
fn with_split(self) -> RegistrationBuilder<...> { ... }
fn with_split_unserializable(self) -> RegistrationBuilder<...> { ... }
```

The first one would require `Response::Item` to be serializable whereas the second one wouldn't. In principle this is more rigorous than the current approach for the (admittedly very weird) case where `Response` is not serializable but `Response::Item` is serializable or vice-versa. With this approach we could just have `RegistrationBuilder<Request, Response, Streams>`.

src/diagram/node_registry.rs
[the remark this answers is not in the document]
koonpeng: ce17e0d6e5c013bb09dbe7b76829fcc3ff49505c

src/diagram/join.rs
koonpeng: Technically this will never fail because we are converting a (Small)Vec of Value to a Value. But I made it `cancel_on_err` just in case.

koonpeng:
It is hard to explain how the `unwraps` in `workflow_builder.rs` are safe. We usually access `edges` with a key that comes from `vertices` and both are built using the same data, so in theory it would never panic. But we do modify `edges` while building the workflow so the preconditions because very complex.

I added a `UnknownDiagramError`, for these unwraps, it captures the backtrace to help debug if it ever happens.

src/diagram.rs
```diff
@@ -472,6 +472,11 @@ impl Diagram {
     }
 }
 
+// thiserror has a special impl for backtrace which only works on nightly.
+// But we can disable this special impl by aliasing it.
+// See https://github.com/dtolnay/thiserror/issues/387#issuecomment-2473781561.
+type DefinitelyNotBacktrace = backtrace::Backtrace;
```
mxgrey, line 478: Just making sure I understand, without this type alias the code will fail to compile on nightly? I assume it should compile without any issue on stable, right?

Also while this name is honestly very funny, we should rename it to something like `BacktraceAlias`.

koonpeng: It's the other way round, it won't compile without this alias, `thiserror` only supports backtrace on nightly.

mxgrey: I'm a little bit concerned about only including the file and line location since users won't always know which version of the source code to compare against. What if we add a `&str` message inside of `unknown_diagram_error!(~)` similar to `.expect(~)` to help narrow down which line had an error?

mxgrey: On the other hand if we're only using this error type for truly unexpected errors, then I don't mind leaving it as-is for now and we can see if any of these errors actually occurs after exercising the feature for a while.

diagram.schema.json
koonpeng: 046f81f99ed7a6b8a7e9143ebc951b5c4ad70336

src/diagram/node_registry.rs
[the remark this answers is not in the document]
mxgrey: Resolved by #45

mxgrey (approved):
Thanks for all the rigorous iterating @koonpeng 

I'm super happy with how this capability turned out. The JSON syntax is very clean and intuitive without limiting what can be done with it. The fact that the schema can be generated from native Rust structs is even better.

We still have plenty more operations to add, but this initial set is already very powerful, and most importantly it sets a strong foundation for us to continue building off of.

PR #30, opened 2024-11-13
Add derive macro for DeliveryLabel

mxgrey:
This PR adds the ability for downstream users to `#[derive(DeliveryLabel, ...)]` on their custom structs.

Unfortunately this derive macro cannot be used inside of `bevy_impulse` itself because I still have not figured out the dark ritual to make that work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy_impulse"
-version = "0.0.1"
+version = "0.0.2"
 edition = "2021"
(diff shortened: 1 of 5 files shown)


luca-della-vedova (approved):
https://users.rust-lang.org/t/proc-macros-using-third-party-crate/42465/4 seems to be a good resource but it does look a bit hairy (it only tackles external dependencies that you seem to have already done with `DynEq`), but something for another time

PR #34, opened 2024-11-27
WIP: Fix 1.75 build failure / add CI

luca-della-vedova:
`take_if` is only stable in Rust 1.80 so removed it inspired on its [implementation in the standard library](https://doc.rust-lang.org/src/core/option.rs.html#1758-1760).

I don't think that is really covered by tests since I had a first implementation that was wrong and it was not caught by `cargo test` so big :warning: there.
I also added a few steps to the Linux CI to install and try Rust 1.75, happy to put it in a different workflow if that is a better option

### DIFF
--- a/.github/workflows/ci_linux.yaml
+++ b/.github/workflows/ci_linux.yaml
@@ -31,3 +31,11 @@ jobs:
       run: cargo build --features single_threaded_async
     - name: Test single_threaded_async
       run: cargo test --features single_threaded_async
+
+    # Build and test with 1.75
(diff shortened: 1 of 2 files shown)


mxgrey:
We could definitely do better in terms of how comprehensive the test coverage is. There are so many different possible combinations of things that people can do with this library that there are nearly uncountable nooks and crannies to consider.

But I've added in a unit test for the array split key limit, so at least that one is covered now.

mxgrey (approved):

PR #35, opened 2024-11-27
Implement Split for json_serde::Value

mxgrey:
This PR implements the new split operation for `serde_json::Value`. I also cleaned up two more things while working on that:
* Give explicit lifetimes to `Chain::split` operations so the `FnOnce` build functions could output something that utilizes those lifetimes
* Refactor `SplitAsMap::next` so that anything using `MapSplitKey` can take advantage of the premade `next` for it.


koonpeng (changes requested):
Is it possible to move the changes in `src/chain.rs` and `src/chain/split.rs` to a separate PR targeting main?

PR #45, opened 2025-01-08
Introduce CommonOperations vs RegistrationBuilder

mxgrey:
This PR is a proposal to address [this discussion thread](https://github.com/open-rmf/bevy_impulse/pull/27#discussion_r1900561042).

It started out with me shifting some code around as an intellectual exercise to understand the technical challenges of the API recommendations that I was making, but eventually it wound up as a working solution, so I decided to commit and push it.

A few notes:
* `gen` appears to be [a reserved keyword](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/gen-keyword.html) starting in 2024, so I replaced it with `schema_generator`.
* Grouping data-related impls into their own `DataRegistry` helps `RegistrationBuilder` to contain simultaneous mutable borrows of a single entry of `NodeRegistry::nodes` and everything it needs to update data-specific impl information.
* We probably have an opportunity to simplify the trait-based implementations of enabling unzip, split, and the others, but we can save that for a future PR since it wouldn't break this API 

I'm very open to renaming things like `CommonOperations` and `fn opt_out`. It's tricky to think of names for these concepts since it's unusual to have an opt-out builder pattern rather than an opt-in builder pattern.

### DIFF
--- a/src/diagram/node_registry.rs
+++ b/src/diagram/node_registry.rs
@@ -664,7 +664,10 @@ pub struct NodeBuilderOptions {
 
 impl NodeBuilderOptions {
     pub fn new(id: impl ToString) -> Self {
-        Self { id: id.to_string(), name: None }
+        Self {
+            id: id.to_string(),
+            name: None,
+        }
     }
 
     pub fn with_name(mut self, name: impl ToString) -> Self {
@@ -672,8 +675,6 @@ impl NodeBuilderOptions {
         self
     }
 }
-
-
 
 #[cfg(test)]
 mod tests {
@@ -693,8 +694,7 @@ mod tests {
             .opt_out()
             .no_response_cloning()
             .register_node_builder(
-                NodeBuilderOptions::new("multiply3_uncloneable")
-                    .with_name("Test Name"),
+                NodeBuilderOptions::new("multiply3_uncloneable").with_name("Test Name"),
                 |builder, _config: ()| builder.create_map_block(multiply3),
             );
         let registration = registry.get_registration("multiply3_uncloneable").unwrap();
@@ -708,9 +708,7 @@ mod tests {
     fn test_register_cloneable_node() {
         let mut registry = NodeRegistry::default();
         registry.register_node_builder(
-            NodeBuilderOptions::new("multiply3")
-                .with_name("Test Name")
-            ,
+            NodeBuilderOptions::new("multiply3").with_name("Test Name"),
             |builder, _config: ()| builder.create_map_block(multiply3),
         );
         let registration = registry.get_registration("multiply3").unwrap();
@@ -728,8 +726,7 @@ mod tests {
             .opt_out()
             .no_response_cloning()
             .register_node_builder(
-                NodeBuilderOptions::new("multiply3_uncloneable")
-                    .with_name("Test Name"),
+                NodeBuilderOptions::new("multiply3_uncloneable").with_name("Test Name"),
                 move |builder: &mut Builder, _config: ()| builder.create_map_block(tuple_resp),
             )
             .with_unzip();
@@ -746,8 +743,7 @@ mod tests {
         let vec_resp = |_: ()| -> Vec<i64> { vec![1, 2] };
         registry
             .register_node_builder(
-                NodeBuilderOptions::new("vec_resp")
-                    .with_name("Test Name"),
+                NodeBuilderOptions::new("vec_resp").with_name("Test Name"),
                 move |builder: &mut Builder, _config: ()| builder.create_map_block(vec_resp),
             )
             .with_split();
@@ -757,8 +753,7 @@ mod tests {
         let map_resp = |_: ()| -> HashMap<String, i64> { HashMap::new() };
         registry
             .register_node_builder(
-                NodeBuilderOptions::new("map_resp")
-                    .with_name("Test Name"),
+                NodeBuilderOptions::new("map_resp").with_name("Test Name"),
                 move |builder: &mut Builder, _config: ()| builder.create_map_block(map_resp),
             )
             .with_split();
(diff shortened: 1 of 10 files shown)


src/diagram/node_registry.rs
```diff
@@ -469,17 +464,20 @@ pub trait IntoNodeRegistration {
...
+        schema_generator: &mut SchemaGenerator,
     ) -> NodeRegistration;
 }
 
 pub struct NodeRegistry {
     nodes: HashMap<BuilderId, NodeRegistration>,
+    pub(super) data: DataRegistry,
+}
 
+pub struct DataRegistry {
```
koonpeng, line 476: I'm inclined to avoid calling it `DataRegistry`, "Data" is too generic. Maybe `CommonOperations` can be renamed to `CommonOperationsBuilder`, then this can be `CommonOperations`.

src/diagram/node_registry.rs
```diff
@@ -700,13 +702,14 @@ mod tests {
...
-            .with_unzippable()
+            .opt_out()
+            .no_response_cloning()
             .register_node_builder(
                 "multiply3_uncloneable".to_string(),
                 "Test Name".to_string(),
                 move |builder: &mut Builder, _config: ()| builder.create_map_block(tuple_resp),
-            );
+            )
+            .with_unzip();
```
koonpeng, line 712: Not important, but I feel that the 2 step builder does result in a bit of "weirdness" in the api. Normally with a builders, I am expecting a pattern where I create the builder, choose my options, then execute with those options. But here you choose some options, executes, then choose more options.

I think we can achieve the usual builder pattern if we add `CommonOperations` as a generic param to `RegistrationBuilder`. It does add one more param but it does not increase with more operations. Or we can defer creating node registration until all the options are selected, something like this could work

```rust
impl CommonOperations<...> {
  fn into_registration() -> {...}

  fn opt_in() -> RegistrationBuilder<...> {...}
}

impl RegistrationBuilder<...> {
  fn register_node_builder(...) {
    let mut reg = self.data.into_registration()?;
    // set opt in fields
  }
}
```

Not sure if this is worth exploring.

src/diagram/node_registry.rs
mxgrey: I named it "data" to refer to the data that moves between nodes. I would actually recommend that we eventually move the impls for `unzip`, `split`, and all the rest into here while `NodeRegistration` only contains a `TypeId` for `Request` and `Response`. That would allow us to have one central database of all operations that can be done on each data type. Then we can register operations for known data types separately from registering node builders. However this may conflict with the node metadata, so we'd need to reconsider how that's presented.

In any case, all of the above ideas can be saved for a future PR.

What if we named this `MessageRegistry` instead of `DataRegistry`?

src/diagram/node_registry.rs
mxgrey: I get what you mean about it being an unusual pattern. Opt-out builders are a pretty unusual thing to support.

>  It does add one more param

I've experimented with this kind of "[policy-based design](https://blog.andreiavram.ro/cpp-policy-based-design/)" using Rust generics in the past, and what ends up happening is you have to repeatedly write out

```rust
impl<DeserializeImpl, SerializeImpl, Cloneable, Request, Response, Streams> RegistrationBuilder<CommonOperations<DeserializeImpl, SerializeImpl, Cloneable>, Request, Response, Streams>
```

all over the place, so you really end up with 6+ complex generic arguments when you thought you'd only have 4 simple ones.

> Or we can defer creating node registration until all the options are selected, something like this could work

I can see this working if we require them to always finish the whole chain with `.build()`. We can put `#[must_use]` onto `RegistrationBuilder` so they don't just drop it accidentally. But I believe this will require the same awkward 6-parameter pattern that I described above.

Personally I think the need to opt-out of serialization, deserialization, and cloning for a diagram node is niche enough that we shouldn't worry too much about how nice the ergonomics are for that, as long as
1. serialization, deseralization, and cloning are enabled by default since most users would expect that
2. they **can** be disabled when it's necessary **_somehow_** without polluting the hot path API
3. the API scales well for adding more opt-in operations in the future

I think the design in the current PR hits all of those points.

Personally I hope this this API is just a stop-gap measure until the Rust language team stabilizes `min_specialization`, at which point we should throw this all out and just use nice clean specialization to automatically infer all relevant operations.

mxgrey:
I've updated this PR to include the change for `NodeBuilderOptions`. I took the liberty of tweaking the API of `NodeBuilderOptions` so that we don't need so much `.to_string()`, and also made `.with_name(~)` more idiomatic for the builder pattern.

src/diagram/node_registry.rs
koonpeng: > all over the place, so you really end up with 6+ complex generic arguments when you thought you'd only have 4 simple ones.

Couldn't we do

```rust
struct RegistratinBuilder<CommonOps, ...> {
  common_ops: CommonOps,
}

impl<CommonOps, ...> RegistrationBuilder<CommonOps, ...> where CommonOps: IntoRegistration {
}
```

This is assuming that we can remove the dependency on `Serializer` and `Deserializer` in `RegistrationBuilder`, which is part of the proposal to move everything into `CommonOps`. Then again, if we can do that without bloating `CommonOps`, then we wouldn't need `CommonOps`, we should be able to do the same on `RegistrationBuilder`.

tbh, I feel that there are many things we can explore to see how we can improve the api and code cleanliness. Unless you are very oppose to the current api, I would prefer to create an issue for this and leave refactoring this for another PR.

koonpeng (approved):

src/diagram/node_registry.rs
mxgrey: > assuming that we can remove the dependency on Serializer and Deserializer in RegistrationBuilder

I don't see how to do that if the user is going to decide to opt out of serializing and deserializing after they have already started building the registration. The only reason it works in this PR is because the decision to opt-out is made before the rest of the registration building takes place.

> Unless you are very oppose to the current api, I would prefer to create an issue for this and leave refactoring this for another PR.

When I think about the other optional operations that I expect us to be adding very soon, I'd like to make sure the first draft of the registration builder is something that will scale well to more operations. I'm very happy to keep iterating on the implementation, but I would recommend the approach in this PR as the starting point.

PR #46, opened 2025-01-09
Merge main into 0.1 release branch

luca-della-vedova:
This PR just brings in the latest changes from `main` into the `release-0.1` branch. Specifically, to unblock downstream migrations on `rmf_site` we need the `DeliveryLabel` macro PR https://github.com/open-rmf/bevy_impulse/pull/30, which is not released.

Eventually I'd like to bring the change all the way to the `release-0.2` branch for Bevy 0.14 migration, I thought the sequence would be a `main` -> `0.1, followed by a `0.1 -> 0.2` but I might have missed something

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustdocflags = ["-D", "warnings"]
--- a/.github/workflows/ci_linux.yaml
+++ b/.github/workflows/ci_linux.yaml
@@ -16,17 +16,10 @@ env:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        rust-version: [stable, 1.75]
-
     runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v3
-
-    - name: Setup rust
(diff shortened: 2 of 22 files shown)


mxgrey:
Looks like we'll need to remove the 1.75 test for this branch or else the CI will always be red.

I anticipate we'll need to keep targeting important features at `v0.0.x` for them to be hypothetically build farm compatible.

luca-della-vedova:
Agreed! I suspect this PR should not be squash merged but normal merged to preserve history? If that is the case should I amend / force push the merge commit to avoid minor changes (specifically to fix the style) or create a `Fix Style` commit?
Removing the CI should still be a separate commit to keep track of the change

mxgrey:
Yeah, regular commit makes sense. The hard part will be remembering to switch the mode when the time comes.

mxgrey (approved):

mxgrey:
Closing this in favor of https://github.com/open-rmf/bevy_impulse/pull/68

PR #50, opened 2025-01-16
Message registry

koonpeng:
### Implemented feature

Move operation impl management to `MessageRegistry`.

The move of impls is actually quite straightforward, the bigger change is in the need to move some node metadatas to messages, particularly flags that indicate if a node is serializable, fork cloneable etc.

The json for "NodeRegistry" (renamed to just "Registry", open to suggestions) now looks like this

```json
{
  "nodes": {
    "test": {
      "metadata": {
        "id": "test",
        "name": "test",
        "request": "bevy_impulse::diagram::registry::tests::test_serialize_registry::Opaque",
        "response": "bevy_impulse::diagram::registry::tests::test_serialize_registry::Bar",
        "config_schema": {
          "type": "null"
        }
      }
    }
  },
  "messages": {
    "bevy_impulse::diagram::registry::tests::test_serialize_registry::Bar": {
      "schema": {
        "$ref": "#/schemas/Bar"
      },
      "operations": [
        "serialize",
        "fork_clone"
      ],
      "unzip_slots": 0
    },
    "bevy_impulse::diagram::registry::tests::test_serialize_registry::Opaque": {
      "schema": null,
      "operations": [],
      "unzip_slots": 0
    }
  },
  "schemas": {
    "Bar": {
      "type": "object",
      "required": [
        "foo"
      ],
      "properties": {
        "foo": {
          "$ref": "#/schemas/Foo"
        }
      }
    },
    "Foo": {
      "type": "object",
      "required": [
        "hello"
      ],
      "properties": {
        "hello": {
          "type": "string"
        }
      }
    }
  }
}
```

There are some minor catches to this

1. The rust `std::any::type_name::<T>()` is used to key the messages, `TypeId` is still used as the key, type name is only used when serializing. The catch here is that type name and `TypeId` do not have a 1-to-1 relationship, i.e. type name is not unique, multiple types may have the same type name. In practice, conflicts shouldn't be common, but if it does, there isn't a solution atm (we will first need either derive or a system to register messages independent of nodes).
2. The rust type name is different from the json schema type name. This shouldn't cause any issue in practice, just a small confusion.

### DIFF
--- a/src/diagram/registry.rs
+++ b/src/diagram/registry.rs
@@ -106,7 +106,9 @@ where
 pub(super) struct NodeMetadata {
     pub(super) id: BuilderId,
     pub(super) name: String,
+    /// type name of the request
     pub(super) request: &'static str,
+    /// type name of the response
     pub(super) response: &'static str,
     pub(super) config_schema: Schema,
 }
--- a/src/diagram/registry.rs
+++ b/src/diagram/registry.rs
@@ -16,7 +16,7 @@ use schemars::{
 };
 use serde::{
     de::DeserializeOwned,
-    ser::{SerializeMap, SerializeStruct},
+    ser::{SerializeMap, SerializeSeq, SerializeStruct},
     Serialize,
 };
 use tracing::debug;
@@ -468,6 +468,14 @@ impl MessageOperation {
         f(builder, output)
     }
 
+    pub(super) fn unzip_slots(&self) -> usize {
+        if let Some(unzip_impl) = &self.unzip_impl {
+            unzip_impl.slots
+        } else {
+            0
+        }
+    }
+
     #[cfg(test)]
     pub(super) fn can_fork_result(&self) -> bool {
         self.fork_result_impl.is_some()
@@ -526,28 +534,34 @@ impl Serialize for MessageOperation {
     where
         S: serde::Serializer,
     {
-        let mut s = serializer.serialize_struct("MessageOperation", 7)?;
-        s.serialize_field("deserialize", &self.deserialize_impl.is_some())?;
-        s.serialize_field("serialize", &self.serialize_impl.is_some())?;
-        s.serialize_field("fork_clone", &self.fork_clone_impl.is_some())?;
-        let unzip_slots = if let Some(unzip_impl) = &self.unzip_impl {
-            unzip_impl.slots
-        } else {
-            0
-        };
-        s.serialize_field("unzip_slots", &unzip_slots)?;
-        s.serialize_field("fork_result", &self.fork_result_impl.is_some())?;
-        s.serialize_field("split", &self.split_impl.is_some())?;
-        s.serialize_field("join", &self.join_impl.is_some())?;
+        let mut s = serializer.serialize_seq(None)?;
+        if self.deserialize_impl.is_some() {
+            s.serialize_element("deserialize")?;
+        }
+        if self.serialize_impl.is_some() {
+            s.serialize_element("serialize")?;
+        }
+        if self.fork_clone_impl.is_some() {
+            s.serialize_element("fork_clone")?;
+        }
+        if self.unzip_impl.is_some() {
+            s.serialize_element("unzip")?;
+        }
+        if self.fork_result_impl.is_some() {
+            s.serialize_element("fork_result")?;
+        }
+        if self.split_impl.is_some() {
+            s.serialize_element("split")?;
+        }
+        if self.join_impl.is_some() {
+            s.serialize_element("join")?;
+        }
         s.end()
     }
 }
 
-#[derive(Serialize)]
 struct MessageRegistration {
-    #[serde(skip)]
     type_name: &'static str,
-
     schema: Option<schemars::schema::Schema>,
     operations: MessageOperation,
 }
@@ -559,6 +573,19 @@ impl MessageRegistration {
             schema: None,
             operations: MessageOperation::new(),
         }
+    }
+}
+
+impl Serialize for MessageRegistration {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut s = serializer.serialize_struct("MessageRegistration", 3)?;
+        s.serialize_field("schema", &self.schema)?;
+        s.serialize_field("operations", &self.operations)?;
+        s.serialize_field("unzip_slots", &self.operations.unzip_slots())?;
+        s.end()
     }
 }
 
--- a/src/diagram/registry.rs
+++ b/src/diagram/registry.rs
@@ -1,8 +1,8 @@
 use std::{
-    any::{Any, TypeId},
+    any::{type_name, Any, TypeId},
     borrow::Borrow,
     cell::RefCell,
-    collections::{hash_map::Entry, HashMap},
+    collections::HashMap,
     fmt::Debug,
     marker::PhantomData,
 };
@@ -14,10 +14,14 @@ use schemars::{
     schema::Schema,
     JsonSchema,
 };
-use serde::{de::DeserializeOwned, ser::SerializeStruct, Serialize};
+use serde::{
+    de::DeserializeOwned,
+    ser::{SerializeMap, SerializeStruct},
+    Serialize,
+};
 use tracing::debug;
 
-use crate::{RequestMetadata, SerializeMessage};
+use crate::SerializeMessage;
 
 use super::{
     fork_clone::DynForkClone,
@@ -25,8 +29,7 @@ use super::{
     impls::{DefaultImpl, NotSupported},
     unzip::DynUnzip,
     BuilderId, DefaultDeserializer, DefaultSerializer, DeserializeMessage, DiagramError, DynSplit,
-    DynSplitOutputs, DynType, OpaqueMessageDeserializer, OpaqueMessageSerializer, ResponseMetadata,
-    SplitOp,
+    DynSplitOutputs, DynType, OpaqueMessageDeserializer, OpaqueMessageSerializer, SplitOp,
 };
 
 /// A type erased [`crate::InputSlot`]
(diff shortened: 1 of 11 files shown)


mxgrey:
> The json for "NodeRegistry" (renamed to just "Registry", open to suggestions) now looks like this

This is looking really good to me! 

> (renamed to just "Registry", open to suggestions)

I'm okay with this, but one idiosyncrasy of Rust is that you usually import objects by their most basic name instead of by their fully qualified name, so the name `Registry` has a high likelihood of colliding with some other `Registry` type. I might suggest something more explicit, like `DiagramElementRegistry`? But I do agree that `NodeRegistry` is no longer the right name for this.

> The rust std::any::type_name::<T>() is used to key the messages when serializing, TypeId is still used as the key in other places.

I think this is the right move.

> multiple types may have the same type name

I think these edge cases are niche enough that we shouldn't worry. If we ever do run into a problem here, we can do something similar to the `name` field for node, where users can optionally register a custom name for each message type (e.g. a UUID), and we'll just use that inside the schema, with `type_name` as the automatic fallback.

> The rust type name is different from the json schema type name. This shouldn't cause any issue in practice, just a small confusion.

I assume you're referring to `Bar` vs `bevy_impulse::diagram::registry::tests::test_serialize_registry::Bar`. I just tested this and found that JsonSchema is smart enough to create a `Bar` and a `Bar2` automatically if two structs share the same name, so I guess we're covered for name collisions.

> What if the user wants to have a unzippable and uncloneable response, but have the unzipped elements be cloneable.

Very interesting point that I hadn't accounted for. In fact the `.with_unzip_unserializable` only indicates that the output is not serializable, but it doesn't indicate that the output will also not be cloneable. I would suggest the following:

* We come up with a better name for `.with_unzip_unserializable`. Something that indicates that it's opting out of all the default operations for the unzipped elements.
* We provide a convenience struct like the following `MessageRegistrationBuilder`:

```rust
pub struct MessageRegistrationBuilder<'a, T> {
    data: &'a MessageRegistry,
    _ignore: std::marker::PhantomData<T>,
}

impl<'a, T> MessageRegistrationBuilder<'a, T> {
    pub fn with_unzip(self) -> Self { ... }
    pub fn with_fork_result(self) -> Self { ... }
}

impl<'a, DeserializeImpl, SerializeImpl, Cloneable> CommonOperations<'a, DeserializeImpl, SerializeImpl, Cloneable> {
    pub fn register_message<T>(self) -> MessageRegistrationBuilder<'a, T> { ... }
}
```

Then users have the power to register operations for each element from the unzip individually.

mxgrey (commented):
I like this refactor a lot. I think it's more logical and will also be more versatile both for registering operations and also for a GUI editor to make sense of what operations are supported and where.

I couldn't find anything in the individual lines of code to iterate on. I would just suggest addressing the few pieces of feedback that I left in my [earlier comment](https://github.com/open-rmf/bevy_impulse/pull/50#issuecomment-2597516121):

* Consider the name `DiagramElementRegistry` or something similar instead of `Registry`.
* Let's try to think of a more accurate name for `.with_unzip_unserializable`
* Add a `MessageRegistrationBuilder` and perhaps rename `RegistrationBuilder` to `NodeRegistrationBuilder`.

mxgrey:
Oh one additional thought: I recommend changing `"operations": [ ... ]` from an array into a dictionary so we can choose to express a configuration for each operation:

```json
"operations": {
    "fork_clone": null,
    "serialize": null,
    "unzip": ["Foo", "Bar"],
    "split": {
        "element_type": "Bar",
        "known_keys": ["foo", "bar"],
        "sequence_maximum": 10
    }
}
```

I don't expect us to have meaningful configuration information per operation yet, but it would be good for us to design the format so we can add it in the future. I think a GUI editor will want to use this information to infer what kind of connections can be made to the outputs of each operation.

koonpeng:
> Oh one additional thought: I recommend changing `"operations": [ ... ]` from an array into a dictionary so we can choose to express a configuration for each operation:
> 
> ```json
> "operations": {
>     "fork_clone": null,
>     "serialize": null,
>     "unzip": ["Foo", "Bar"],
>     "split": {
>         "element_type": "Bar",
>         "known_keys": ["foo", "bar"],
>         "sequence_maximum": 10
>     }
> }
> ```
> 
> I don't expect us to have meaningful configuration information per operation yet, so we can leave them all as `null` for now, but it would be good for us to design the format so we can add it in the future. I think a GUI editor will want to use this information to infer what kind of connections can be made to the outputs of each operation.

82a53f211d22ae4cff7a89a6584ab17150633b3c

src/diagram/registry.rs
```diff
@@ -282,64 +302,137 @@ impl<'a, DeserializeImpl, SerializeImpl, Cloneable>
...
+        DefaultImpl: DynSplit<Message, DefaultSerializer>,
+    {
+        self.data
+            .register_split::<Message, DefaultImpl, DefaultSerializer>();
+        self
+    }
+
+    /// Mark the node as having a splittable response but the items from the split
+    /// are unserializable.
+    pub fn with_split_unserializable(self) -> Self
```
mxgrey, line 363: Same here, let's replace `unserializable` with a better term.

src/diagram/registry.rs
```diff
@@ -282,64 +302,137 @@ impl<'a, DeserializeImpl, SerializeImpl, Cloneable>
...
         DefaultImplMarker<(Response, DefaultSerializer)>: DynUnzip,
     {
-        self.data.register_unzip::<Response, DefaultSerializer>();
+        MessageRegistrationBuilder::new(self.registry).with_unzip();
         self
     }
 
     /// Mark the node as having an unzippable response whose elements are not serializable.
-    pub fn with_unzip_unserializable(self) -> Self
+    pub fn with_unzip_unserializable(&mut self) -> &mut Self
```
mxgrey, line 401: Replace unserializable

src/diagram/registry.rs
```diff
@@ -282,64 +302,137 @@ impl<'a, DeserializeImpl, SerializeImpl, Cloneable>
...
-        self.data
-            .register_split::<Response, DefaultImpl, DefaultSerializer>();
+        MessageRegistrationBuilder::new(self.registry).with_split();
         self
     }
 
     /// Mark the node as having a splittable response but the items from the split
     /// are unserializable.
-    pub fn with_split_unserializable(self) -> Self
+    pub fn with_split_unserializable(&mut self) -> &mut Self
```
mxgrey, line 431: Replace unserializable

src/diagram/registry.rs
```diff
@@ -0,0 +1,1421 @@
...
+type SplitFn = Box<
+    dyn for<'a> Fn(
+        &mut Builder,
+        DynOutput,
+        &'a SplitOp,
+    ) -> Result<DynSplitOutputs<'a>, DiagramError>,
+>;
+type JoinFn = Box<dyn Fn(&mut Builder, Vec<DynOutput>) -> Result<DynOutput, DiagramError>>;
+
+pub struct CommonOperations<'a, Deserialize, Serialize, Cloneable> {
```
mxgrey, line 189: Unrelated to this PR but we should probably add `#[must_use]` to this struct.

mxgrey (changes requested):
This PR is looking very good, I just have a few more nits to pick, mostly related to naming.

src/diagram/registry.rs
```diff
@@ -0,0 +1,1421 @@
...
+    fn into_node_registration(
+        self,
+        id: BuilderId,
+        name: String,
+        schema_generator: &mut SchemaGenerator,
+    ) -> NodeRegistration;
+}
+
+#[derive(Serialize)]
+pub struct DiagramElementRegistry {
```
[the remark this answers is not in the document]
koonpeng: 6cf280988df4a468bce851037a41e430e9227ab7...4e6d8ac6f8c2fed0de962a38ab56016f72634eb2

src/diagram/registry.rs
```diff
@@ -282,64 +302,137 @@ impl<'a, DeserializeImpl, SerializeImpl, Cloneable>
...
+    pub fn with_unzip(self) -> Self
+    where
+        DefaultImplMarker<(Message, DefaultSerializer)>: DynUnzip,
+    {
+        self.data.register_unzip::<Message, DefaultSerializer>();
+        self
+    }
+
+    /// Mark the node as having an unzippable response whose elements are not serializable.
+    pub fn with_unzip_unserializable(self) -> Self
```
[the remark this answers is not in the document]
koonpeng: 6cf280988df4a468bce851037a41e430e9227ab7..4e6d8ac6f8c2fed0de962a38ab56016f72634eb2

src/diagram/registry.rs
koonpeng: 6cf280988df4a468bce851037a41e430e9227ab7..4e6d8ac6f8c2fed0de962a38ab56016f72634eb2

src/diagram/registry.rs
koonpeng: 6cf280988df4a468bce851037a41e430e9227ab7..4e6d8ac6f8c2fed0de962a38ab56016f72634eb2

src/diagram/registry.rs
koonpeng: 6cf280988df4a468bce851037a41e430e9227ab7..4e6d8ac6f8c2fed0de962a38ab56016f72634eb2

src/diagram/registry.rs
koonpeng: 6cf280988df4a468bce851037a41e430e9227ab7..4e6d8ac6f8c2fed0de962a38ab56016f72634eb2

src/diagram/registry.rs
```diff
@@ -0,0 +1,1421 @@
...
+            s.serialize_entry("deserialize", &serde_json::Value::Null)?;
+        }
+        if self.serialize_impl.is_some() {
+            s.serialize_entry("serialize", &serde_json::Value::Null)?;
+        }
+        if self.fork_clone_impl.is_some() {
+            s.serialize_entry("fork_clone", &serde_json::Value::Null)?;
+        }
+        if let Some(unzip_impl) = &self.unzip_impl {
+            s.serialize_entry("unzip", &json!({"slots": unzip_impl.slots()}))?;
```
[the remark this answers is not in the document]
koonpeng: 6cf280988df4a468bce851037a41e430e9227ab7...4e6d8ac6f8c2fed0de962a38ab56016f72634eb2

mxgrey (approved):
Awesome, I think this is an excellent redesign of the operation registration system :tada:

PR #52, opened 2025-02-02
Support for dynamic dictionaries of buffers

mxgrey:
This PR is to address issues in the core library that are creating roadblocks for #51

This is a hefty PR but adds some important layers of flexibility to buffers. The highlights are:
* `AnyBuffer` can represent a buffer of any message type, and is now a first-class citizen of the library
  * Use `AnyBufferKey` to access any buffer through the `World`
* `JsonBuffer` can represent a buffer of any message type that supports serialization and deserialization
  * Use `JsonBufferKey` to access the JSON data in the buffer through the `World`
* `BufferMap` can now be used with `Builder::join_into` to join a dictionary of `AnyBuffer` instances into a value that implements `JoinedValue`
  * This should make it much easier and more robust to define how the join operation works in JSON and convert that into actual workflow connections

I'm leaving this PR as a draft because there are a few more items to finish:
- [x] Create a `Joined` macro that implements the `JoinedValue` and `BufferMapLayout` traits for any struct
- [x] Create a `BufferKeyMap` macro that implements the `BufferKeyMap` and `BufferMapLayout` traits for any struct that consists entirely of `BufferKey<T>`,  and/or `AnyBufferKey` fields
- [x] Implement `Builder::try_listen` to convert a `BufferMap` into a specific struct that implements `BufferKeyMap`.

### DIFF
--- a/macros/src/buffer.rs
+++ b/macros/src/buffer.rs
@@ -57,7 +57,7 @@ pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream
     let impl_joined = impl_joined(&buffer_struct, &input_struct, &field_ident)?;
 
     let gen = quote! {
-        impl #impl_generics ::bevy_impulse::JoinedValue for #struct_ident #ty_generics #where_clause {
+        impl #impl_generics ::bevy_impulse::Joined for #struct_ident #ty_generics #where_clause {
             type Buffers = #buffer_struct_ident #ty_generics;
         }
 
@@ -122,7 +122,7 @@ pub(crate) fn impl_buffer_key_map(input_struct: &ItemStruct) -> Result<TokenStre
     let impl_accessed = impl_accessed(&buffer_struct, &input_struct, &field_ident, &field_type)?;
 
     let gen = quote! {
-        impl #impl_generics ::bevy_impulse::BufferKeyMap for #struct_ident #ty_generics #where_clause {
+        impl #impl_generics ::bevy_impulse::Accessor for #struct_ident #ty_generics #where_clause {
             type Buffers = #buffer_struct_ident #ty_generics;
         }
 
@@ -413,8 +413,8 @@ fn impl_buffer_map_layout(
 }
 
 /// Params:
-///   joined_struct: The struct to implement `Joined`.
-///   item_struct: The associated `Item` type to use for the `Joined` implementation.
+///   joined_struct: The struct to implement `Joining`.
+///   item_struct: The associated `Item` type to use for the `Joining` implementation.
 fn impl_joined(
     joined_struct: &ItemStruct,
     item_struct: &ItemStruct,
@@ -425,7 +425,7 @@ fn impl_joined(
     let (impl_generics, ty_generics, where_clause) = item_struct.generics.split_for_impl();
 
     Ok(quote! {
-        impl #impl_generics ::bevy_impulse::Joined for #struct_ident #ty_generics #where_clause {
+        impl #impl_generics ::bevy_impulse::Joining for #struct_ident #ty_generics #where_clause {
             type Item = #item_struct_ident #ty_generics;
 
             fn pull(&self, session: ::bevy_impulse::re_exports::Entity, world: &mut ::bevy_impulse::re_exports::World) -> Result<Self::Item, ::bevy_impulse::OperationError> {
@@ -452,7 +452,7 @@ fn impl_accessed(
     let (impl_generics, ty_generics, where_clause) = key_struct.generics.split_for_impl();
 
     Ok(quote! {
-        impl #impl_generics ::bevy_impulse::Accessed for #struct_ident #ty_generics #where_clause {
+        impl #impl_generics ::bevy_impulse::Accessing for #struct_ident #ty_generics #where_clause {
             type Key = #key_struct_ident #ty_generics;
 
             fn add_accessor(
@@ -461,7 +461,7 @@ fn impl_accessed(
                 world: &mut ::bevy_impulse::re_exports::World,
             ) -> ::bevy_impulse::OperationResult {
                 #(
-                    ::bevy_impulse::Accessed::add_accessor(&self.#field_ident, accessor, world)?;
+                    ::bevy_impulse::Accessing::add_accessor(&self.#field_ident, accessor, world)?;
                 )*
                 Ok(())
             }
--- a/macros/src/buffer.rs
+++ b/macros/src/buffer.rs
@@ -1,58 +1,60 @@
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
-use syn::{parse_quote, Field, Generics, Ident, ItemStruct, Type, TypePath};
+use syn::{
+    parse_quote, Field, Generics, Ident, ImplGenerics, ItemStruct, Type, TypeGenerics, TypePath,
+    Visibility, WhereClause,
+};
 
 use crate::Result;
+
+const JOINED_ATTR_TAG: &'static str = "joined";
+const KEY_ATTR_TAG: &'static str = "key";
 
 pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream> {
     let struct_ident = &input_struct.ident;
     let (impl_generics, ty_generics, where_clause) = input_struct.generics.split_for_impl();
     let StructConfig {
         buffer_struct_name: buffer_struct_ident,
-    } = StructConfig::from_data_struct(&input_struct);
+    } = StructConfig::from_data_struct(&input_struct, &JOINED_ATTR_TAG);
     let buffer_struct_vis = &input_struct.vis;
 
-    let (field_ident, _, field_config) = get_fields_map(&input_struct.fields)?;
+    let (field_ident, _, field_config) =
+        get_fields_map(&input_struct.fields, FieldSettings::for_joined())?;
     let buffer: Vec<&Type> = field_config.iter().map(|config| &config.buffer).collect();
     let noncopy = field_config.iter().any(|config| config.noncopy);
 
-    let buffer_struct: ItemStruct = parse_quote! {
-        #[allow(non_camel_case_types, unused)]
-        #buffer_struct_vis struct #buffer_struct_ident #impl_generics #where_clause {
-            #(
-                #buffer_struct_vis #field_ident: #buffer,
-            )*
-        }
-    };
+    let buffer_struct: ItemStruct = generate_buffer_struct(
+        &buffer_struct_ident,
+        buffer_struct_vis,
+        &impl_generics,
+        &where_clause,
+        &field_ident,
+        &buffer,
+    );
 
-    let impl_buffer_clone = if noncopy {
-        // Clone impl for structs with a buffer that is not copyable
-        quote! {
-            impl #impl_generics ::std::clone::Clone for #buffer_struct_ident #ty_generics #where_clause {
-                fn clone(&self) -> Self {
-                    Self {
-                        #(
-                            #field_ident: self.#field_ident.clone(),
-                        )*
-                    }
-                }
-            }
-        }
-    } else {
-        // Clone and copy impl for structs with buffers that are all copyable
-        quote! {
-            impl #impl_generics ::std::clone::Clone for #buffer_struct_ident #ty_generics #where_clause {
-                fn clone(&self) -> Self {
-                    *self
-                }
-            }
+    let impl_buffer_clone = impl_buffer_clone(
+        &buffer_struct_ident,
+        &impl_generics,
+        &ty_generics,
+        &where_clause,
+        &field_ident,
+        noncopy,
+    );
 
-            impl #impl_generics ::std::marker::Copy for #buffer_struct_ident #ty_generics #where_clause {}
-        }
-    };
+    let impl_select_buffers = impl_select_buffers(
+        struct_ident,
+        &buffer_struct_ident,
+        buffer_struct_vis,
+        &impl_generics,
+        &ty_generics,
+        &where_clause,
+        &field_ident,
+        &buffer,
+    );
 
-    let impl_buffer_map_layout = impl_buffer_map_layout(&buffer_struct, &input_struct)?;
-    let impl_joined = impl_joined(&buffer_struct, &input_struct)?;
+    let impl_buffer_map_layout =
+        impl_buffer_map_layout(&buffer_struct, &field_ident, &field_config)?;
+    let impl_joined = impl_joined(&buffer_struct, &input_struct, &field_ident)?;
 
     let gen = quote! {
         impl #impl_generics ::bevy_impulse::JoinedValue for #struct_ident #ty_generics #where_clause {
@@ -63,23 +65,76 @@ pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream
 
         #impl_buffer_clone
 
-        impl #impl_generics #struct_ident #ty_generics #where_clause {
-            fn select_buffers(
-                #(
-                    #field_ident: #buffer,
-                )*
-            ) -> #buffer_struct_ident #ty_generics {
-                #buffer_struct_ident {
-                    #(
-                        #field_ident,
-                    )*
-                }
-            }
-        }
+        #impl_select_buffers
 
         #impl_buffer_map_layout
 
         #impl_joined
+    };
+
+    Ok(gen.into())
+}
+
+pub(crate) fn impl_buffer_key_map(input_struct: &ItemStruct) -> Result<TokenStream> {
+    let struct_ident = &input_struct.ident;
+    let (impl_generics, ty_generics, where_clause) = input_struct.generics.split_for_impl();
+    let StructConfig {
+        buffer_struct_name: buffer_struct_ident,
+    } = StructConfig::from_data_struct(&input_struct, &KEY_ATTR_TAG);
+    let buffer_struct_vis = &input_struct.vis;
+
+    let (field_ident, field_type, field_config) =
+        get_fields_map(&input_struct.fields, FieldSettings::for_key())?;
+    let buffer: Vec<&Type> = field_config.iter().map(|config| &config.buffer).collect();
+    let noncopy = field_config.iter().any(|config| config.noncopy);
+
+    let buffer_struct: ItemStruct = generate_buffer_struct(
+        &buffer_struct_ident,
+        buffer_struct_vis,
+        &impl_generics,
+        &where_clause,
+        &field_ident,
+        &buffer,
+    );
+
+    let impl_buffer_clone = impl_buffer_clone(
+        &buffer_struct_ident,
+        &impl_generics,
+        &ty_generics,
+        &where_clause,
+        &field_ident,
+        noncopy,
+    );
+
+    let impl_select_buffers = impl_select_buffers(
+        struct_ident,
+        &buffer_struct_ident,
+        buffer_struct_vis,
+        &impl_generics,
+        &ty_generics,
+        &where_clause,
+        &field_ident,
+        &buffer,
+    );
+
+    let impl_buffer_map_layout =
+        impl_buffer_map_layout(&buffer_struct, &field_ident, &field_config)?;
+    let impl_accessed = impl_accessed(&buffer_struct, &input_struct, &field_ident, &field_type)?;
+
+    let gen = quote! {
+        impl #impl_generics ::bevy_impulse::BufferKeyMap for #struct_ident #ty_generics #where_clause {
+            type Buffers = #buffer_struct_ident #ty_generics;
+        }
+
+        #buffer_struct
+
+        #impl_buffer_clone
+
+        #impl_select_buffers
+
+        #impl_buffer_map_layout
+
+        #impl_accessed
     };
 
     Ok(gen.into())
@@ -112,7 +167,7 @@ struct StructConfig {
 }
 
 impl StructConfig {
-    fn from_data_struct(data_struct: &ItemStruct) -> Self {
+    fn from_data_struct(data_struct: &ItemStruct, attr_tag: &str) -> Self {
         let mut config = Self {
             buffer_struct_name: format_ident!("__bevy_impulse_{}_Buffers", data_struct.ident),
         };
@@ -120,7 +175,7 @@ impl StructConfig {
         let attr = data_struct
             .attrs
             .iter()
-            .find(|attr| attr.path().is_ident("joined"));
+            .find(|attr| attr.path().is_ident(attr_tag));
 
         if let Some(attr) = attr {
             attr.parse_nested_meta(|meta| {
@@ -137,23 +192,52 @@ impl StructConfig {
     }
 }
 
+struct FieldSettings {
+    default_buffer: fn(&Type) -> Type,
+    attr_tag: &'static str,
+}
+
+impl FieldSettings {
+    fn for_joined() -> Self {
+        Self {
+            default_buffer: Self::default_field_for_joined,
+            attr_tag: JOINED_ATTR_TAG,
+        }
+    }
+
+    fn for_key() -> Self {
+        Self {
+            default_buffer: Self::default_field_for_key,
+            attr_tag: KEY_ATTR_TAG,
+        }
+    }
+
+    fn default_field_for_joined(ty: &Type) -> Type {
+        parse_quote! { ::bevy_impulse::Buffer<#ty> }
+    }
+
+    fn default_field_for_key(ty: &Type) -> Type {
+        parse_quote! { <#ty as ::bevy_impulse::BufferKeyLifecycle>::TargetBuffer }
+    }
+}
+
 struct FieldConfig {
     buffer: Type,
     noncopy: bool,
 }
 
 impl FieldConfig {
-    fn from_field(field: &Field) -> Self {
+    fn from_field(field: &Field, settings: &FieldSettings) -> Self {
         let ty = &field.ty;
         let mut config = Self {
-            buffer: parse_quote! { ::bevy_impulse::Buffer<#ty> },
+            buffer: (settings.default_buffer)(ty),
             noncopy: false,
         };
 
         for attr in field
             .attrs
             .iter()
-            .filter(|attr| attr.path().is_ident("joined"))
+            .filter(|attr| attr.path().is_ident(settings.attr_tag))
         {
             attr.parse_nested_meta(|meta| {
                 if meta.path.is_ident("buffer") {
@@ -172,7 +256,10 @@ impl FieldConfig {
     }
 }
 
-fn get_fields_map(fields: &syn::Fields) -> Result<(Vec<&Ident>, Vec<&Type>, Vec<FieldConfig>)> {
+fn get_fields_map(
+    fields: &syn::Fields,
+    settings: FieldSettings,
+) -> Result<(Vec<&Ident>, Vec<&Type>, Vec<FieldConfig>)> {
     match fields {
         syn::Fields::Named(data) => {
             let mut idents = Vec::new();
@@ -185,7 +272,7 @@ fn get_fields_map(fields: &syn::Fields) -> Result<(Vec<&Ident>, Vec<&Type>, Vec<
                     .ok_or("expected named fields".to_string())?;
                 idents.push(ident);
                 types.push(&field.ty);
-                configs.push(FieldConfig::from_field(field));
+                configs.push(FieldConfig::from_field(field, &settings));
             }
             Ok((idents, types, configs))
         }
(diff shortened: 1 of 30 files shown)


koonpeng (commented):
I feel there is quite a lot of duplicated code between `Buffer`, `AnyBuffer` and `JsonBuffer`. In particular, I'm not sure how useful `JsonBuffer` would be over `Buffer<serde_json::Value>` (possibly with conversion to `AnyBuffer`).

I can see how `AnyBuffer` would be useful for diagrams, but it doesn't address the issue that diagrams can only create `Buffer<AnyMessage>` and converting that to `AnyBuffer` loses a lot of the advantages of `AnyBuffer`.

`join_into` seems like to would help as well, but I don't see how I can use it in diagrams because `buffer_access`, `listen` and `join` do not know the buffers or buffer maps at compile time. We can register join impls but the issue is that the workflow builder does not do any forward lookups atm, so it doesn't know which impl to use. Thats not a hard blocker but if it does support forward lookups, then `join` should work as well as `join_into`.

src/buffer/any_buffer.rs
```diff
@@ -0,0 +1,1169 @@
...
+    }
+
+    fn any_message_type(&self) -> TypeId {
+        TypeId::of::<T>()
+    }
+}
+
+pub type AnyMessageMut<'a> = &'a mut (dyn Any + 'static + Send + Sync);
+
+pub type AnyMessage = Box<dyn Any + 'static + Send + Sync>;
```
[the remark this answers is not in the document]
mxgrey: The use of Box can't be helped because we need a container for the vtable of the trait.

> maybe rename it to AnyMessageRefMut?

It's currently following the pattern in bevy where the `Ref` suffix refers to something that resembles a shared borrow while the `Mut` suffix refers to something that resembles a mutable borrow. If we're going to rename anything I think I'd rather rename `AnyMessage` to `AnyMessageBox`.

> Also I think the definition for all the AnyMessage aliases should be together.

Sure, happy to reorganize this.

src/buffer/any_buffer.rs
```diff
@@ -0,0 +1,1169 @@
...
+        world: &mut World,
+    ) -> Box<dyn AnyBufferAccessMutState>;
+}
+
+pub(crate) struct AnyBufferAccessImpl<T>(std::marker::PhantomData<T>);
+
+impl<T: 'static + Send + Sync + Any> AnyBufferAccessImpl<T> {
+    pub(crate) fn get_interface() -> &'static (dyn AnyBufferAccessInterface + Send + Sync) {
+        const INTERFACE_MAP: OnceLock<
+            Mutex<HashMap<TypeId, &'static (dyn AnyBufferAccessInterface + Send + Sync)>>,
```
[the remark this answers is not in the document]
mxgrey: This is a pattern of using a global variable to cache static references, so the mutex can't be helped. I don't think there's a need to document it since it'll be invisible to the user and the amount of actual blocking it does should be negligible, nearly unobservable.

mxgrey:
> I feel there is quite a lot of duplicated code between Buffer, AnyBuffer and JsonBuffer.

I agree and I'm happy to iterate on some aspects of the implementation, now that we have an MVP. Those three structs in particular should be reasonable to refactor.

But I think most of the duplication between the modules (in terms of lines of code) comes from the various `BufferAccess...` traits. Initially I tried to funnel all those traits into one, but I ran into limitations in Rust's generics system that forced me to keep them all separate. I think the best thing we can do there to cut down noise is move those traits into sub-submodules so they're at least not eating up real estate within the modules that we expect users to be looking at.

> I can see how `AnyBuffer` would be useful for diagrams, but it doesn't address the issue that diagrams can only create `Buffer<AnyMessage>`

For diagrams every message type would auto-register a vtable to create a `AnyBuffer` for the appropriate message type. We would **not** use `Buffer<AnyMessage>` at all.

> join_into seems like to would help as well, but I don't see how I can use it in diagrams

Due to the tragic absence of specialization we'll need to introduce a registration function along the lines of `.with_joined` which can be used on any message types that implement the `JoinedValue` trait. That will register a callback that can insert a join element into the workflow given a `BufferMap`.

I haven't finished the pieces that will be needed for buffer access and listening, but they will work pretty similarly to `join_into`.

> the issue is that the workflow builder does not do any forward lookups atm

You're right that this will probably require forward lookups. We'll probably need a two-stage pass over all the operations where the first pass identifies the message types for all operations where that's fixed and then the second pass infers the message types for the rest based on their connections.

> Thats not a hard blocker but if it does support forward lookups, then join should work as well as join_into.

That's true, we don't strictly need `join_into`, but something I realized while thinking all of this over is that using tuples for all the joins, buffer access, and listening is much too fragile. If a user puts the wrong ordering of elements into the tuples we'll be introducing confusing bugs that are hard to identify.

The real value of this new approach is that we can assign explicit names to each item that is part of a join or part of a group of buffers. Now the json format can use explicit dictionaries of buffers instead of fragile arrays of buffers. That's the key advantage that I'm trying to unlock with this PR.

mxgrey:
> I'm not sure how useful `JsonBuffer` would be over `Buffer<serde_json::Value>`

Forgot to comment on this one. There's a very important difference: `JsonBuffer` is driven by some underlying concrete data type, which means insertions are protected by a schema. If someone tries to insert a nonsense or incompatible `serde_json::Value` it will be rejected. All insertions must be able to deserialize into the underlying type. In contrast `Buffer<serde_json::Value>` would unquestioningly accept any serialized message that someone pushes into it.

There's also the fact that `JsonBuffer` makes serialization an **option** and not a **requirement** for the buffer. If I have a `Buffer<Pose>` that I convert to a `JsonBuffer` I can still use it as a `Buffer<Pose>` throughout the workflow. But I additionally have the **option** to interact with it as if it contains `serde_json::Value` content.

src/buffer/any_buffer.rs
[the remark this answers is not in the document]
koonpeng: Didn't know about the bevy convention, but that makes sense. I think `AnyMessageBox` makes sense as well, technically `AnyMessage` is a `dyn Any + Send...`.

src/buffer/buffer_map.rs
```diff
@@ -0,0 +1,456 @@
...
+        assert_eq!(value.string, "hello");
+        assert!(context.no_unhandled_errors());
+    }
+
+    #[test]
+    fn test_joined_value() {
+        let mut context = TestingContext::minimal_plugins();
+
+        let workflow = context.spawn_io_workflow(|scope, builder| {
+            let buffers = TestJoinedValueBuffers {
```
koonpeng, line 428: While implementing #53, I noticed a problem with the new api here. Normally code generated from a `derive` should be "implementation detail" and users shouldn't ever need to interact with it directly. After we move this to derive, this here will be using a generated struct directly.

Not sure how feasible is it to remove the need for this struct, but I guess there are at least some ways to avoid explicit use of the struct.
* Have a `TestJoinedValue::new_buffers` to create the struct, not ideal as users still need to reference this struct if they use it in function arguments etc.
* Do not auto generate this struct, instead have the user define and derive both `TestJoinedValue` and `TestJoinedValueBuffers`.

mxgrey: Yeah this was also an issue before the recent refactor to the implementation of `JoinedValue`, but now it's more obvious.

This problem is the main reason that I originally only used tuples for collections of buffers, because I was worried that the use of some invisibly defined struct would create more confusion than benefit. But ultimately I've concluded that nodes will **need** to be able to name the buffers that they're ingesting, otherwise 
it will be too confusing for users to make sure they're providing the correct buffers into the correct slots, especially when sketching diagrams.

> Not sure how feasible is it to remove the need for this struct

I don't see any way to avoid it if we want to be able to have a name for each buffer and allow the compiler to validate everything.

Generating the `_Buffers` struct has the additional advantage of allowing us to generate a custom `Joined` implementation for the generated struct which makes the joining operation as efficient as it can possibly be.

> Have a `TestJoinedValue::new_buffers` to create the struct

I think this is a good idea, but I might suggest naming the function something like `JoinedValue::select_buffers(...)` or `JoinedValue::collect_buffers(...)` instead since "new_buffers" sounds like you're creating new buffers when in reality you're putting together a collection of existing buffers that will be joined into the value.

Out of curiosity, I tested a different approach that I thought could work well:

```rust
let buffers = TestJoinedValue::Buffers {
    integer: builder.create_buffer(BufferSettings::default()),
    float: builder.create_buffer(BufferSettings::default()),
    string: builder.create_buffer(BufferSettings::default()),
};
```

however this failed because the syntax [is not supported yet](https://github.com/rust-lang/rust/issues/86935). Even in the [unstable implementation](https://github.com/rust-lang/rust/pull/80080) of the language feature it looks like the syntax would be `let buffers = <TestJoinedValue as JoinedValue>::Buffers {` which is still rather ugly and not something that users are likely to want.

> Do not auto generate this struct, instead have the user define and derive both `TestJoinedValue` and `TestJoinedValueBuffers`.

I expect users will quickly get annoyed by the duplication of maintenance between the two structs if we do this. In almost all cases I think users will primarily care about handling the `JoinedValue` struct, and the `_Buffers` struct is just a technicality of the implementation.

---

On a slightly different but related note, something I wanted to ask for is a `macro_rules!` that we would call `buffers!{...}`. Its usage would look something like:

```rust
let join = builder.try_join(buffers!{
    "integer": buffer_i64,
    "float": buffer_f64,
    "string": buffer_string,
});
```

The `buffers!` macro would produce a `BufferMap` struct by just calling `.insert(~,~)` on each comma-separated row. Then users can pass that `BufferMap` into `try_join`.

koonpeng: What I'm thinking of is something like this

```rust
let buffers = TestJoinedValue::new_buffers();

scope.input.chain(builder).fork_unzip((
    |chain: Chain<_>| chain.connect(buffers.integer.input_slot()),
    |chain: Chain<_>| chain.connect(buffers.float.input_slot()),
    |chain: Chain<_>| chain.connect(buffers.string.input_slot()),
));

builder.join(buffers).connect(scope.terminate);
```

This way, the `_Buffers` struct can be inferred and never needs to appear in a `use`, but you still need to `use` it if you have it in a function argument or return etc.

But `select_buffers` work as well, I guess it would look like

```rust
let buffer_i64 = builder.create_buffer(BufferSettings::default());
let buffer_f64 = builder.create_buffer(BufferSettings::default());
let buffer_string = builder.create_buffer(BufferSettings::default());

scope.input.chain(builder).fork_unzip((
    |chain: Chain<_>| chain.connect(buffer_i64.input_slot()),
    |chain: Chain<_>| chain.connect(buffer_f64.input_slot()),
    |chain: Chain<_>| chain.connect(buffer_string.input_slot()),
));

builder.join(TestJoinedValue::select_buffers(buffer_i64, buffer_f64, buffer_string)).connect(scope.terminate);
```

I like that it reduces the scope of the derive options, with `new_buffers` we need to have options for the buffer types and settings, with `select_buffers` it can take `Bufferable<T>` and that should automatically support all kinds of buffers. Actually this might be a really important detail as `new_buffers` approach have a hard time supporting joining between different types of buffers (e.g. `Output<T>`, `CloneFromBuffer<T>` etc).

---

It seems that with `try_join`, we don't need a `_Buffers` struct at all since the buffers are stored in a `BufferMap`. If we are ok with only supporting `try_join` for derived buffers, then I guess we don't need to generate any new structs.

---

> On a slightly different but related note, something I wanted to ask for is a macro_rules! that we would call buffers!{...}. Its usage would look something like:

I'm not sure if we should have that, for the same reason why there is no `hashmap!` macro in rust. The "canonical" way to create a hashmap is `from_iter`. There is the `maplit` crate, which does that for a bunch of std structs, but the inconvenience for all these alternatives is that you probably have to call `.into()` on the buffers. I can see how `buffers!` make things more convenient, but coming from c++, I'm adverse to abusing macros like this.

mxgrey: > let buffers = TestJoinedValue::new_buffers();

Ah I see what you're getting at. I can see this as a convenience function, but it has some drawbacks:
1. It doesn't give an opportunity to customize the buffer settings per buffer
2. It gets in the way of some of the other flexible ways that we allow users to create buffers

I could see us supporting something like

```
let buffers = TestJoinedValue::create_buffers(builder, buffer_settings);
```

1. We'll need to pass in the `builder` when creating buffers
2. `create_buffers` matches existing terminology better
3. The `buffer_settings` can apply to all the buffers that get created

But I think we'll still need a `select_buffers` function for situations that need more flexibility, and you already listed some of the reasons for that.

> If we are ok with only supporting try_join for derived buffers, then I guess we don't need to generate any new structs.

When you say "for derived buffers" do you mean only supporting it for structs that implement `JoinedValue`? Due to how Rust traits work, we have no choice but to only support it for `JoinedValue` types.

You're right that we could theoretically implement all this without generating a `_Buffers` struct (and in fact `try_join_into` was originally implemented like that via `BufferedMap<JoinedValue>`), but we lose two benefits of the current implementation:
1. There would be no way to support `builder.join<JoinedValue>(buffers)`, so users would always need to use `try_join` to get a `JoinedValue`, and `try_join` has runtime fallibility whereas `join` doesn't.
2. The current implementation generates a `OperateJoin<_Buffers>` per `_Buffers` struct instead of using hashmap lookups, which means more efficient machine code and less runtime fallibility.

I think we should just do our best to give ways for users to not need to think about the name of the `_Buffers` struct. With some usage examples of `select_buffers` in the doctests of `JoinedValue`, I think users won't have too much trouble understanding how to use it.

---

> for the same reason why there is no hashmap! macro in rust. The "canonical" way to create a hashmap is from_iter.

I feel that `from_iter` is rather noisy for this use case. It also means we'd need to make this kind of API available to `BufferMap`. At that point we'd probably be better off just making `BufferMap` an alias for `HashMap<Cow<'static, str>, AnyBuffer>`, but I'm iffy about committing to that. I could probably be convinced to go with that, but I'd rather `BufferMap` be a wrapper so we have opportunities to customize the API and tweak the implementation without breaking the API.

> coming from c++, I'm adverse to abusing macros like this.

I wouldn't consider this  to be an abuse of macros. Plenty of popular libraries provide builder macros, like `smallvec![_]` from `smallvec`. Tbh I don't see a particularly compelling rationale for the absence of `hashmap!` in Rust. Just some ambiguity about what the syntax should look like is the best explanation I've seen [[1]](https://internals.rust-lang.org/t/map-macro-as-a-hashmap-counter-of-vec/16720/4). But then there are crates like `maplit` and also [`velcro`](https://crates.io/crates/velcro).

An advantage of a `buffers!` macro is it can automatically apply `.as_any_buffer()` to each buffer which is not something Rust's builtin mechanisms would do, so I think it could _**really**_ cut down on noise.

On the other hand, if we try out `select_buffers` and find it to be adequate then maybe `buffers!` doesn't have much of a use case anyway. For the `diagram` implementation of buffers we'll just use `BufferMap` directly anyway.

mxgrey:
@koonpeng I've added support for using `builder.join(...)` to go from `Vec<Buffer>` -> `Vec<Value>`.

I've also added explicit support for `HashMap<String, JsonBuffer>` -> `serde_json::Map<String, serde_json::Value>` and `Vec<JsonBuffer>` -> `Vec<serde_json::Value>`. I think to utilize these in the diagram we'll probably need to introduce a `serialize_join` operation which is separate from `join`.
* Users would choose between providing a `name: buffer` dictionary or an array of buffers
* We cast the buffers into `JsonBuffer`s, and join them using `HashMap<String, JsonBuffer>` or `Vec<JsonBuffer>` depending on what the user chose
* The overall operation would output a `serde_json::Value`

The advantage of this `serialize_join` is it can work for any input buffers whose message types have serialization registered, and the diagram could deserialize the joined output into any message type that has deserialization registered, even if the final output doesn't have the join operation registered.

Now the only thing left before I take this PR out of draft is introducing keys structs for the `listen` and `with_access` operations.

src/buffer/buffer_map.rs
```diff
@@ -24,27 +24,66 @@ use smallvec::SmallVec;
...
+    }
+}
+
+pub type BufferMap = HashMap<BufferIdentifier<'static>, AnyBuffer>;
+
+/// Extension trait that makes it more convenient to insert buffers into a [`BufferMap`].
+pub trait AddBufferToMap {
+    /// Convenience function for inserting items into a [`BufferMap`]. This
+    /// automatically takes care of converting the types.
+    fn insert_buffer<I: Into<BufferIdentifier<'static>>, B: AsAnyBuffer>(
```
koonpeng, line 73: I think a potential problem with this is that index based keys are not guaranteed to be sequential, I don't know if this is good or not in practice, but I think we can change `insert_buffer` to only accept the string variant, and have a `extend_buffer` that takes an iterator and inserts the index keys sequentially.

I also thought of making `BufferMap` an enum

```rust
pub enum BufferMap<'a> {
  String(HashMap<Cow<'a, str>, AnyBuffer>),
  Index(Vec<AnyBuffer>),
}
```

But I guess there is probably a reason for the current impl.

macros/src/buffer.rs
```diff
@@ -208,19 +208,17 @@ fn impl_buffer_map_layout(
...
-                )*]
-            }
-
             fn try_from_buffer_map(buffers: &::bevy_impulse::BufferMap) -> Result<Self, ::bevy_impulse::IncompatibleLayout> {
                 let mut compatibility = ::bevy_impulse::IncompatibleLayout::default();
                 #(
-                    let #field_ident = if let Ok(buffer) = compatibility.require_buffer_type::<#buffer>(#map_key, buffers) {
-                        buffer
-                    } else {
+                    let #field_ident = compatibility.require_buffer_by_literal::<#buffer>(#map_key, buffers);
```
koonpeng, line 214: Is there a reason why we can't use `require_buffer_by_name`? I tried changing to that and the tests still passes.

src/buffer/buffer_map.rs
```diff
@@ -53,57 +92,93 @@ impl BufferMap {
...
         }
 
         Err(())
     }
 
-    /// Check whether a named buffer is compatible with a specialized buffer type,
-    /// such as `JsonBuffer`.
-    pub fn require_buffer_type<BufferType: 'static>(
+    /// Check whether a named buffer is compatible with the required buffer type.
+    pub fn require_buffer_by_literal<BufferType: 'static>(
```
koonpeng, line 146: Is there a reason why we couldn't use `require_buffer_by_name`, passing it a `&'static str` literal? It doesn't seem to break anything when I tried that.

src/buffer/buffer_map.rs
```diff
@@ -53,57 +92,93 @@ impl BufferMap {
...
+        buffers: &BufferMap,
+    ) -> Result<BufferType, ()> {
+        self.require_buffer::<BufferType>(BufferIdentifier::Index(expected_index), buffers)
+    }
+
+    /// Check whether the buffer associated with the identifier is compatible with
+    /// the required buffer type.
+    pub fn require_buffer<BufferType: 'static>(
+        &mut self,
+        identifier: BufferIdentifier<'static>,
```
koonpeng, line 167: Could this take `impl Into<BufferIdentifier<'static>>` and support both name and index keys? Assuming we impl `From<usize> for BufferIdentifier<'static>`.

src/buffer/buffer_map.rs
```diff
@@ -311,9 +393,49 @@ impl Accessed for BufferMap {
...
+
+        compatibility.as_result()?;
+        Ok(downcast_buffers)
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use crate::{prelude::*, testing::*, BufferMap};
+    use crate::{prelude::*, testing::*, AddBufferToMap, BufferMap};
```
koonpeng, line 438: Should `AddBufferToMap` be in prelude?

macros/src/buffer.rs
mxgrey: By name will do a heap allocation if an error occurs while by literal will not, since it can do a permanent borrow of the static memory. 

It's very minor but I figure we may as well take advantage of the savings.

src/buffer/buffer_map.rs
mxgrey: Even if the borrow that you pass in has a static lifetime, the function can't take advantage of that, so it will need to do a heap allocation if an error occurs.

src/buffer/buffer_map.rs
mxgrey: I'm a bit torn on this and open to opinions. I'm leaning towards no because realistically I think it'll only be used in tests and maybe in the diagram module. I think it's very unlikely that downstream users will ever use a `BufferMap` directly.

src/buffer/buffer_map.rs
mxgrey: My expectation is that no one will do ordered iteration using this key. If someone wants to iterate over indices then they will `.get(&BufferIdentifier::Index(i))`.

Alternatively they may match the key to `BufferIdentifier::Index(index)` and then use `index` to align with the layout.

mxgrey: > But I guess there is probably a reason for the current impl.

Mostly I found that funneling everything into the key made it much easier to have a simple and consistent error message. This also theoretically supports maps that can have a mix of names and indices but I don't think we're likely to need that.

koonpeng:
> I've also added explicit support for HashMap<String, JsonBuffer> -> serde_json::Map<String, serde_json::Value> and Vec<JsonBuffer> -> Vec<serde_json::Value>. I think to utilize these in the diagram we'll probably need to introduce a serialize_join operation which is separate from join.

Currently, to do this, you need to join multiple outputs with the serialize flag on, it would produce a `Vec<Value>`, then you can put this to a `Transform` op to do any CEL transforms. I would prefer to have this feature through `Transform`, but I like the ability to specify a map.

```jsonc
{
  type: "transform",
  cel: "{ \"hello\": node_a, \"world\": node_b.msg }",
  next: ...
}
```

So `Transform` would do an implicit `serialized_join`, with all the keys based on the id of the node, e.g. "node_a" would be referenced as "node_a" in the CEL etc. The CEL is actually very similar to a `name: buffer` dict in `serialized_join`, the biggest problem is that you need to escape it (not an issue with a UI). The advantage of `Transform` is that you can build a `Value` from fields of an output, e.g. if `node_b` outputs a `Foo { msg: String }` struct, `serialized_join` cannot construct the value.

mxgrey:
> So `Transform` would do an implicit `serialized_join`

The problem with `Transform` is it doesn't actually do joining. `Transform` is an operation that requires a single input and then transforms that input into some output. I'd rather not have a situation where `Transform` sometimes takes an input and sometimes performs a join depending on context.

> So `Transform` would do an implicit `serialized_join`, with all the keys based on the id of the node, e.g. "node_a" would be referenced as "node_a" in the CEL etc.

If I understand the proposal correctly, we would need to parse the CEL string, identify any node or buffer references that exist inside the string, build up the buffer map based on those references, and then serialize+join the values from the buffers before passing them to the `Transform` operation. I think we'd end up doing the `serialized_join` anyway before transforming.

> The advantage of `Transform` is that you can build a `Value` from fields of an output, e.g. if `node_b` outputs a `Foo { msg: String }` struct, `serialized_join` cannot construct the value.

You're right, about this advantage of `Transform`, but I still don't think we can get good join semantics for `Transform` without doing some convoluted things in the implementation. What I would propose is that users take a two-step approach in these situations, e.g.:

```json
{
    "joining": {
        "type": "serialize_join",
        "buffers": {
            "foo": "buffer_a",
            "bar": "buffer_b"
        },
        "next": "transforming"
    },
    "transforming": {
        "type": "transform",
        "cel": "{ \"hello\": foo, \"world\": bar.msg }",
        "next": { "builtin": "terminate" }
    }
}
```

If we're concerned about user convenience we can have widgets in the GUI that represent both of these operations combined.

macros/src/buffer.rs
koonpeng: I feel that this is a very small optimization that makes the api confusing. There isn't any way to avoid passing a `&'static str` to `require_buffer_by_name`, the best we can do is take a `&String`, but imo that creates more downsides than it's worth. I guess the best can be done is to have more docs.

koonpeng:
Just to confirm if my understanding is correct, `join` op would now look like this

```jsonc
{
  "buffer_a": {
    "type": "buffer"
  },
  "buffer_b": {
    "type": "buffer"
  }
  "join": {
    "type": "join",
    "buffers": {
      "foo": "buffer_a",
      "bar": "buffer_b"
    }
    "next": ...
  },
  "join_vec": {
    "type": "join",
    "buffers": ["buffer_a", "buffer_b"],
    "next": ...
  },
  "serialize_join": {
    "type": "serialize_join",
    "buffers": {
      "foo": "buffer_a",
      "bar": "buffer_b"
    },
    "next": ...
  }
}
```

1. join ops would either have to do forward lookup or have a `output_type` field (would need a type name -> `TypeInfo` map). Forward lookup here shouldn't be hard if we only limit to lookup the next node.

1. Joining multiple outputs would require them to be converted to buffers first (should we auto convert outputs to buffers?).

1. "buffer" ops accepts only a single input, it will convert a `DynOutput` into a `AnyBuffer`. No forward lookups required.
    1. To support multiple inputs, we probably need either a `item_type` field or forward lookups. Forward lookup here would be very hard because the type of the node request is the `JoinedValue`, but each buffer is fields to that, and `std::any::Any` does not provide any relations information.

mxgrey:
> 1. Forward lookup here shouldn't be hard if we only limit to lookup the next node.

The problem here is that the next operation won't necessarily be a node. E.g. `fork_clone`, `unzip`, or `buffer`. Most operations that infer their type do it based on the input type rather than the output type, so if `join` infers from its output and gets connected to an operation that infers from its input, then the intermediate type will be ambiguous. For `fork_clone` we could look one step further to infer the message type, but that won't be the case for other operations like `buffer` or `unzip`.

I don't mind if we make it optional to specify the `output_type` for `join`s that are followed by a `node`, but I don't think we can avoid it for other cases. And since we can't avoid it in those other cases, I'd be inclined to think that we should just always specify it for the sake of consistency.

> 2. Joining multiple outputs would require them to be converted to buffers first (should we auto convert outputs to buffers?).

To be clear, even when using the Rust API, outputs are always converted into buffers when being joined. But the Rust API can do this implicitly via the [`Bufferable`](https://github.com/open-rmf/bevy_impulse/blob/0ee3fe2c8fc73d65f400420fe4b48f249a9922bd/src/buffer/bufferable.rs#L29) trait. The name `Bufferable` was used to describe "things that can be turned into buffers", and that includes both `Output<T>` and all flavors of `Buffer`. The API like this is purely for the sake of convenience because I anticipate that joining a bunch of outputs will be a common use-case, and I wanted a simple API for doing that. But still, conceptually they are always converted into buffers.

When it comes to `diagram`, I think the best thing for consistency is to require outputs to be explicitly turned into buffers before being used in a join operation. My rationale is this:
* Operations that produce an output have some kind of `next`-like field that says where its contents should go next, making them push-like.
* Buffer operations don't have a `next` field. Instead operations that use buffers are pull-like by specifying which buffers they are pulling from.
* If a push-like operation connects to a pull-like operation then we have a multiple source of truth situation. This isn't a problem if both sources of truth are consistent, and we can just emit an error if they are inconsistent, but the incongruity doesn't feel good to me.
* Ultimately this is meant to be an intermediate format between a GUI and the workflow executor, so we don't need to go out of our way to make shortcuts within the format. Human readability is important because we want to make sure that workflows can be reviewed and sanity checked by people, but saving a small number of keystrokes shouldn't be a priority for the format.

> 3. "buffer" ops accepts only a single input, it will convert a DynOutput into a AnyBuffer. No forward lookups required. ... To support multiple inputs, we probably need either a item_type field or forward lookups.

If we do type inference for `buffer` operations, we should do it based on the incoming connections. We should check that all incoming connections have the same message type. I'm not sure if this counts as the "forward lookup" that you're talking about since it would be based on incoming message types, similar to `fork_clone`, which I believe is already supported.

If inferring from the inputs is feasible, I think we should go with that. Otherwise there's nothing wrong with having an `item_type` field for the `buffer` operation. Either way we definitely can't infer from the operations downstream of the buffer because of the reasons that you've mentioned.

mxgrey:
For `serialize_join` we can support both a map of buffers and an array of buffers, and output a `serde_json::Value::Map` or `serde_json::Value::Array` based on whichever is chosen.

src/buffer/any_buffer.rs
```diff
@@ -0,0 +1,1169 @@
...
+        > = OnceLock::new();
+        let binding = INTERFACE_MAP;
+
+        let interfaces = binding.get_or_init(|| Mutex::default());
+
+        let mut interfaces_mut = interfaces.lock().unwrap();
+        *interfaces_mut
+            .entry(TypeId::of::<T>())
+            .or_insert_with(|| Box::leak(Box::new(AnyBufferAccessImpl::<T>(Default::default()))))
```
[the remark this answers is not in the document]
mxgrey: I've added `SAFETY` notes on all leaks: [911a5e1](https://github.com/open-rmf/bevy_impulse/pull/52/commits/911a5e17e9baf77f4df648a943aa567fa3d022bb)

src/buffer/buffer_map.rs
```diff
@@ -24,27 +24,66 @@ use smallvec::SmallVec;
...
+        BufferIdentifier::Name(Cow::Borrowed(name))
+    }
+
+    /// Use an index as an identifier.
+    pub fn index(index: usize) -> Self {
+        BufferIdentifier::Index(index)
+    }
+}
+
+impl From<&'static str> for BufferIdentifier<'static> {
```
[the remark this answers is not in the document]
mxgrey: Good idea, added: [530c92f](https://github.com/open-rmf/bevy_impulse/pull/52/commits/530c92f84eedcf012d588d6ff090b23ab9012dc9)

src/buffer/buffer_map.rs
mxgrey: I like this idea. I've changed it here: [530c92f](https://github.com/open-rmf/bevy_impulse/pull/52/commits/530c92f84eedcf012d588d6ff090b23ab9012dc9)

I changed the name to `require_buffer_by_identifier` in case in the future someone wants a more general `require_buffer` that could represent "This buffer type must exist somewhere in the map".

mxgrey:
Taking this out of the draft stage since all the needed features are implemented and all current feedback has been addressed.

I may circle back to add more documentation in some places.

koonpeng (approved):

PR #53, opened 2025-02-05
derive `Joined` to implement `JoinedValue` and `BufferMapLayout`

koonpeng:
### Status:

Updated to the new `join` api. Added derive helper attributes, but the only config now is to change the generated buffer struct name.

* With `select_buffers` there should no longer be a need to change the buffer type. Allowing to customize the types also lead to other issues as explained in https://github.com/open-rmf/bevy_impulse/pull/53#issuecomment-2652846213.
* It is not possible to change expose pub idents in a private span and vice versa ([E0446](https://doc.rust-lang.org/error_codes/E0446.html)). In derive macros, my guess is that it inherits the visibility of the target struct, attempting to generate structs with different visibility than the original will result in E0446.

---

For now only struct with named fields is supported, tuple struct or newtype struct doesn't make sense as a buffer map layout. enums are questionable but they will require more logic to validate and select which branch to use.

### DIFF
--- a/macros/src/buffer.rs
+++ b/macros/src/buffer.rs
@@ -1,21 +1,24 @@
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
-use syn::{parse_quote, Field, Ident, ItemStruct, Type};
+use syn::{parse_quote, Ident, ItemStruct, Type};
 
 use crate::Result;
 
 pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream> {
     let struct_ident = &input_struct.ident;
     let (impl_generics, ty_generics, where_clause) = input_struct.generics.split_for_impl();
-    let (field_ident, field_type, _) = get_fields_map(&input_struct.fields)?;
-    let struct_buffer_ident = format_ident!("__bevy_impulse_{}_Buffers", struct_ident);
+    let (field_ident, field_type) = get_fields_map(&input_struct.fields)?;
+    let BufferStructConfig {
+        struct_name: buffer_struct_ident,
+    } = BufferStructConfig::from_data_struct(&input_struct);
+    let buffer_struct_vis = &input_struct.vis;
 
     let buffer_struct: ItemStruct = parse_quote! {
         #[derive(Clone)]
         #[allow(non_camel_case_types)]
-        struct #struct_buffer_ident #impl_generics #where_clause {
+        #buffer_struct_vis struct #buffer_struct_ident #impl_generics #where_clause {
             #(
-                #field_ident: ::bevy_impulse::Buffer<#field_type>,
+                #buffer_struct_vis #field_ident: ::bevy_impulse::Buffer<#field_type>,
             )*
         }
     };
@@ -25,7 +28,7 @@ pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream
 
     let gen = quote! {
         impl #impl_generics ::bevy_impulse::JoinedValue for #struct_ident #ty_generics #where_clause {
-            type Buffers = #struct_buffer_ident #ty_generics;
+            type Buffers = #buffer_struct_ident #ty_generics;
         }
 
         #buffer_struct
@@ -36,8 +39,8 @@ pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream
                 #(
                     #field_ident: impl ::bevy_impulse::Bufferable<BufferType = ::bevy_impulse::Buffer<#field_type>>,
                 )*
-            ) -> #struct_buffer_ident #ty_generics {
-                #struct_buffer_ident {
+            ) -> #buffer_struct_ident #ty_generics {
+                #buffer_struct_ident {
                     #(
                         #field_ident: #field_ident.into_buffer(builder),
                     )*
@@ -53,52 +56,50 @@ pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream
     Ok(gen.into())
 }
 
-struct FieldConfig {
-    buffer_type: TokenStream,
+struct BufferStructConfig {
+    struct_name: Ident,
 }
 
-impl FieldConfig {
-    fn from_field(field: &Field) -> Self {
-        let ty = &field.ty;
-        let mut buffer_type = quote! { ::bevy_impulse::Buffer<#ty> };
+impl BufferStructConfig {
+    fn from_data_struct(data_struct: &ItemStruct) -> Self {
+        let mut config = Self {
+            struct_name: format_ident!("__bevy_impulse_{}_Buffers", data_struct.ident),
+        };
 
-        let attr = field
+        let attr = data_struct
             .attrs
             .iter()
-            .find(|attr| attr.path().is_ident("bevy_impulse"));
+            .find(|attr| attr.path().is_ident("buffers"));
 
         if let Some(attr) = attr {
             attr.parse_nested_meta(|meta| {
-                if meta.path.is_ident("buffer_type") {
-                    buffer_type = meta.value()?.parse()?;
+                if meta.path.is_ident("struct_name") {
+                    config.struct_name = meta.value()?.parse()?;
                 }
                 Ok(())
             })
-            // panic if attribute is malformed, this will result in a compile error
+            // panic if attribute is malformed, this will result in a compile error which is intended.
             .unwrap();
         }
 
-        Self { buffer_type }
+        config
     }
 }
 
-fn get_fields_map(fields: &syn::Fields) -> Result<(Vec<&Ident>, Vec<&Type>, Vec<FieldConfig>)> {
+fn get_fields_map(fields: &syn::Fields) -> Result<(Vec<&Ident>, Vec<&Type>)> {
     match fields {
         syn::Fields::Named(data) => {
             let mut idents = Vec::new();
             let mut types = Vec::new();
-            let mut configs = Vec::new();
             for field in &data.named {
                 let ident = field
                     .ident
                     .as_ref()
                     .ok_or("expected named fields".to_string())?;
-                let config = FieldConfig::from_field(field);
                 idents.push(ident);
                 types.push(&field.ty);
-                configs.push(config);
             }
-            Ok((idents, types, configs))
+            Ok((idents, types))
         }
         _ => return Err("expected named fields".to_string()),
     }
@@ -113,7 +114,7 @@ fn impl_buffer_map_layout(
 ) -> Result<proc_macro2::TokenStream> {
     let struct_ident = &buffer_struct.ident;
     let (impl_generics, ty_generics, where_clause) = buffer_struct.generics.split_for_impl();
-    let (field_ident, field_type, _) = get_fields_map(&item_struct.fields)?;
+    let (field_ident, field_type) = get_fields_map(&item_struct.fields)?;
     let map_key: Vec<String> = field_ident.iter().map(|v| v.to_string()).collect();
 
     Ok(quote! {
@@ -154,7 +155,7 @@ fn impl_joined(
     let struct_ident = &joined_struct.ident;
     let item_struct_ident = &item_struct.ident;
     let (impl_generics, ty_generics, where_clause) = item_struct.generics.split_for_impl();
-    let (field_ident, _, _) = get_fields_map(&item_struct.fields)?;
+    let (field_ident, _) = get_fields_map(&item_struct.fields)?;
 
     Ok(quote! {
         impl #impl_generics ::bevy_impulse::Joined for #struct_ident #ty_generics #where_clause {
--- a/macros/src/buffer.rs
+++ b/macros/src/buffer.rs
@@ -1,14 +1,13 @@
-use proc_macro::TokenStream;
+use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
-use syn::{parse_quote, Ident, ItemStruct, Type};
+use syn::{parse_quote, Field, Ident, ItemStruct, Type};
 
 use crate::Result;
 
 pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream> {
     let struct_ident = &input_struct.ident;
     let (impl_generics, ty_generics, where_clause) = input_struct.generics.split_for_impl();
-    let (field_ident, field_type): (Vec<_>, Vec<_>) =
-        get_fields_map(&input_struct.fields)?.into_iter().unzip();
+    let (field_ident, field_type, _) = get_fields_map(&input_struct.fields)?;
     let struct_buffer_ident = format_ident!("__bevy_impulse_{}_Buffers", struct_ident);
 
     let buffer_struct: ItemStruct = parse_quote! {
@@ -33,13 +32,14 @@ pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream
 
         impl #impl_generics #struct_ident #ty_generics #where_clause {
             fn select_buffers(
+                builder: &mut ::bevy_impulse::Builder,
                 #(
-                    #field_ident: ::bevy_impulse::Buffer<#field_type>,
+                    #field_ident: impl ::bevy_impulse::Bufferable<BufferType = ::bevy_impulse::Buffer<#field_type>>,
                 )*
             ) -> #struct_buffer_ident #ty_generics {
                 #struct_buffer_ident {
                     #(
-                        #field_ident,
+                        #field_ident: #field_ident.into_buffer(builder),
                     )*
                 }
             }
@@ -53,18 +53,52 @@ pub(crate) fn impl_joined_value(input_struct: &ItemStruct) -> Result<TokenStream
     Ok(gen.into())
 }
 
-fn get_fields_map(fields: &syn::Fields) -> Result<Vec<(&Ident, &Type)>> {
+struct FieldConfig {
+    buffer_type: TokenStream,
+}
+
+impl FieldConfig {
+    fn from_field(field: &Field) -> Self {
+        let ty = &field.ty;
+        let mut buffer_type = quote! { ::bevy_impulse::Buffer<#ty> };
+
+        let attr = field
+            .attrs
+            .iter()
+            .find(|attr| attr.path().is_ident("bevy_impulse"));
+
+        if let Some(attr) = attr {
+            attr.parse_nested_meta(|meta| {
+                if meta.path.is_ident("buffer_type") {
+                    buffer_type = meta.value()?.parse()?;
+                }
+                Ok(())
+            })
+            // panic if attribute is malformed, this will result in a compile error
+            .unwrap();
+        }
+
+        Self { buffer_type }
+    }
+}
+
+fn get_fields_map(fields: &syn::Fields) -> Result<(Vec<&Ident>, Vec<&Type>, Vec<FieldConfig>)> {
     match fields {
         syn::Fields::Named(data) => {
-            let mut idents_types = Vec::with_capacity(data.named.len());
+            let mut idents = Vec::new();
+            let mut types = Vec::new();
+            let mut configs = Vec::new();
             for field in &data.named {
                 let ident = field
                     .ident
                     .as_ref()
                     .ok_or("expected named fields".to_string())?;
-                idents_types.push((ident, &field.ty));
+                let config = FieldConfig::from_field(field);
+                idents.push(ident);
+                types.push(&field.ty);
+                configs.push(config);
             }
-            Ok(idents_types)
+            Ok((idents, types, configs))
         }
         _ => return Err("expected named fields".to_string()),
     }
@@ -79,8 +113,7 @@ fn impl_buffer_map_layout(
 ) -> Result<proc_macro2::TokenStream> {
     let struct_ident = &buffer_struct.ident;
     let (impl_generics, ty_generics, where_clause) = buffer_struct.generics.split_for_impl();
-    let (field_ident, field_type): (Vec<_>, Vec<_>) =
-        get_fields_map(&item_struct.fields)?.into_iter().unzip();
+    let (field_ident, field_type, _) = get_fields_map(&item_struct.fields)?;
     let map_key: Vec<String> = field_ident.iter().map(|v| v.to_string()).collect();
 
     Ok(quote! {
@@ -121,8 +154,7 @@ fn impl_joined(
     let struct_ident = &joined_struct.ident;
     let item_struct_ident = &item_struct.ident;
     let (impl_generics, ty_generics, where_clause) = item_struct.generics.split_for_impl();
-    let (field_ident, _): (Vec<_>, Vec<_>) =
-        get_fields_map(&item_struct.fields)?.into_iter().unzip();
+    let (field_ident, _, _) = get_fields_map(&item_struct.fields)?;
 
     Ok(quote! {
         impl #impl_generics ::bevy_impulse::Joined for #struct_ident #ty_generics #where_clause {
--- a/macros/src/buffer.rs
+++ b/macros/src/buffer.rs
@@ -6,61 +6,18 @@ use crate::Result;
 
 pub(crate) fn impl_joined_value(ast: DeriveInput) -> Result<TokenStream> {
     let struct_ident = ast.ident;
-    let (field_ident, field_type) = match ast.data {
-        syn::Data::Struct(data) => get_fields_map(data.fields)?,
+    let (field_ident, field_type): (Vec<Ident>, Vec<Type>) = match ast.data {
+        syn::Data::Struct(data) => get_fields_map(data.fields)?.into_iter().unzip(),
         _ => return Err("expected struct".to_string()),
     };
     let map_key: Vec<String> = field_ident.iter().map(|v| v.to_string()).collect();
     let struct_buffer_ident = format_ident!("__{}Buffers", struct_ident);
 
+    let impl_buffer_map_layout =
+        buffer_map_layout(&struct_ident, &field_ident, &field_type, &map_key);
+
     let gen = quote! {
-        impl BufferMapLayout for #struct_ident {
-            fn is_compatible(buffers: &BufferMap) -> Result<(), ::bevy_impulse::IncompatibleLayout> {
-                let mut compatibility = ::bevy_impulse::IncompatibleLayout::default();
-                #(
-                    compatibility.require_buffer::<#field_type>(#map_key, buffers);
-                )*
-                compatibility.into_result()
-            }
-
-            fn buffered_count(
-                buffers: &::bevy_impulse::BufferMap,
-                session: ::bevy_ecs::prelude::Entity,
-                world: &::bevy_ecs::prelude::World,
-            ) -> Result<usize, ::bevy_impulse::OperationError> {
-                use ::bevy_impulse::{InspectBuffer, OrBroken};
-
-                #(
-                    let #field_ident = world
-                        .get_entity(buffers.get(#map_key).or_broken()?.id())
-                        .or_broken()?
-                        .buffered_count::<#field_type>(session)?;
-                )*
-
-                Ok([#( #field_ident ),*]
-                    .iter()
-                    .min()
-                    .copied()
-                    .unwrap_or(0))
-            }
-
-            fn ensure_active_session(
-                buffers: &::bevy_impulse::BufferMap,
-                session: ::bevy_ecs::prelude::Entity,
-                world: &mut ::bevy_ecs::prelude::World,
-            ) -> ::bevy_impulse::OperationResult {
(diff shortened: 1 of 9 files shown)


mxgrey:
The latest version of #52 refactors the way `JoinedValue` is implemented to make it significantly simpler. Here are the highlights:
* Buffer<ins>ed</ins>Map has been removed. BufferMap remains as a generic container of named buffers.
* We no longer need `Builder::join_into`. The `JoinedValue`s can be produced by the already existing `Builder::join`. The new `Builder::try_join_into` has been renamed to simply `Builder::try_join`.
* The `JoinedValue` trait implementation only needs to define the associate type field `type Buffers`.
* The `JoinedValue::Buffers` type now must implement `BufferMapLayout` and `Joined` where the `Joined::Item` must match the `JoinedValue` that it is associated with.
* I've renamed `IncompatibleLayout::require_buffer` into `IncompatibleLayout::require_message_type` and then added `IncompatibleLayout::require_buffer_type`. The latter allows users to require specialized buffer types like `JsonBuffer`. In the future we will likely add more like `ProtoBuffer` and/or `RosBuffer`.

The trait implementation examples for these changes can be [found here](https://github.com/open-rmf/bevy_impulse/blob/858c7e6d11149502e57899153ad113afb9aa33ac/src/buffer/buffer_map.rs#L318-L384). You should find that it's overall much simpler than the previous implementation.

I'd like to request a few additional features for the `JoinedValue` proc macro:
* We should have some attribute that lets the user override how private the `_Buffers` struct is (i.e. `pub`, `pub(crate)`, private). By default the privacy should match the struct that is being derived.
* I recommend all the fields for the `_Buffers` struct should be `pub`. Let the privacy of the struct dictate the overall privacy.
* Provide an attribute that lets the user override the name of the `_Buffers` struct.
* Provide an attribute that lets the user override the expected buffer type for one of the fields in the `JoinedValue`, e.g. the user can apply `#[buffer(JsonBuffer)]` to a field in the `JoinedValue` struct so that the `_Buffers` struct [uses a `JsonBuffer`](https://github.com/open-rmf/bevy_impulse/blob/858c7e6d11149502e57899153ad113afb9aa33ac/src/buffer/json_buffer.rs#L1367) for that field.

koonpeng:
In this implementation, instead of allowing to customize the buffer type, there is a `select_buffer` function that works for any buffer types. I think this is better because of several problems I discovered while trying to implement customizing buffer types.

1. In some cases, it will lead to invalid code being generated.

e.g.

```rust
  #[derive(Clone, JoinedValue)]
  struct TestJoinedValue<T: Send + Sync + 'static + Clone> {
      integer: i64,
      float: f64,
      string: String,
      #[bevy_impulse(buffer_type = AnyBuffer)]
      generic: T,
  }
```

this will generate

```rust
#[derive(Clone)]
#[allow(non_camel_case_types)]
struct __bevy_impulse_TestJoinedValue_Buffers<T: Send + Sync + 'static + Clone> {
    integer: ::bevy_impulse::Buffer<i64>,
    float: ::bevy_impulse::Buffer<f64>,
    string: ::bevy_impulse::Buffer<String>,
    generic: AnyBuffer,
}
```

`T` is never used so we get a compile error. In order to workaround this, we would need to analyze all the fields and remove generics that is never used.

2. `AnyBuffer` does not impl `as_any_buffer`, so this code is invalid.

```rust
  fn buffer_list(&self) -> ::smallvec::SmallVec<[AnyBuffer; 8]> {
      use smallvec::smallvec;
      smallvec![
          self.integer.as_any_buffer(),
          self.float.as_any_buffer(),
          self.string.as_any_buffer(),
          self.generic.as_any_buffer(),
      ]
  }
```

An easy fix is to impl `as_any_buffer`.

3. Different buffers has different pull schematics

```rust
impl<T: Send + Sync + 'static + Clone> ::bevy_impulse::Joined
    for __bevy_impulse_TestJoinedValue_Buffers<T>
{
    type Item = TestJoinedValue<T>;
    fn pull(
        &self,
        session: ::bevy_ecs::prelude::Entity,
        world: &mut ::bevy_ecs::prelude::World,
    ) -> Result<Self::Item, ::bevy_impulse::OperationError> {
        let integer = self.integer.pull(session, world)?;
        let float = self.float.pull(session, world)?;
        let string = self.string.pull(session, world)?;
        let generic = self.generic.pull(session, world)?;
        Ok(Self::Item {
            integer,
            float,
            string,
            generic,
        })
    }
}
```

This fails as `generic` ends up as a `AnyMessageBox`. To get `T`, we need to either downcast it or convert `AnyBuffer` into `Buffer<T>`, I guess the latter is preferred, but we can't do the conversion as we don't have access to a `Builder`.

4. Buffer types are strict and fixed, so only 1 type of buffers can enjoy the convenience.

```rust
// assume `generic_buffer` is a `AnyBuffer`
let user_friendly  = __bevy_impulse_TestJoinedValue_Buffer {
  ...
  generic: generic_buffer,
}
// assume `generic_output` is a `Output<T>`
let not_friendly = __bevy_impulse_TestJoinedValue_Buffer {
  ...
  generic: generic_output.into_buffer().as_any_buffer(),
}
...
```

Because the buffer type is fixed, a conversion needs to be done to use other types of buffers.

5. Special buffer types will need to be converted into `Buffer<T>` anyway when we `pull` from it.

koonpeng:
Tried to implement customizing the generated struct visibility but it seems that it is not possible to change expose pub idents in a private span and vice versa ([E0446](https://doc.rust-lang.org/error_codes/E0446.html)). In derive macros, my guess is that it inherits the visibility of the target struct, attempting to generate structs with different visibility than the original will result in E0446.

mxgrey:
> there is a select_buffer function that works for any buffer types.

Can you show how your approach would work for the [`TestJoinedValueJson`](https://github.com/open-rmf/bevy_impulse/blob/aa22f87fb645b84bdc1d171c24e91c9c09d528dc/src/buffer/json_buffer.rs#L1356-L1368) case?

> 1. In some cases, it will lead to invalid code being generated.

I think this should be easy to fix by adding a field along the lines of `_ignore: ::std::marker::PhantomData<fn( #impl_generics )>`. It might be a little tricky to strip the trait bounds out of `#impl_generics`, but it should be doable by removing all substrings that start with `:` and end with either `,` or `>`.

> 2. AnyBuffer does not impl as_any_buffer, so this code is invalid.

There's no reason we can't implement `as_any_buffer` for `AnyBuffer`. It would just return `self`.

> 3. Different buffers has different pull schematics ... This fails as generic ends up as a AnyMessageBox.

This is exactly why I believe we **_need_** to allow the user to specify the buffer type for special cases. The original use case you showed of

```rust
#[bevy_impulse(buffer_type = "AnyBuffer")]
generic: T,
```

would be invalid and **_should_** produce a compilation error. The use case we **_actually_** want it for would be this:

```rust
#[bevy_impulse(buffer_type = "AnyBuffer")]
any_message: AnyMessageBox,
```

In this example, the `any_message` field is prepared to receive a message of **_any kind_** from _**any**_ buffer. If we aren't able to specialize the buffer type, then it would only be able to receive messages from a `Buffer<AnyMessageBox>` type of buffer.

> 5. Special buffer types will need to be converted into Buffer<T> anyway when we pull from it.

No, this is not the case. `AnyBuffer` can be pulled from to produce a `AnyMessageBox` and `JsonBuffer` can be pulled from to produce a `JsonMessage`. Both have implemented the `Joined` trait so you can use `Joined::pull` on them just like any concrete buffer type.

macros/src/buffer.rs
```diff
@@ -0,0 +1,175 @@
...
+    let gen = quote! {
+        impl #impl_generics ::bevy_impulse::JoinedValue for #struct_ident #ty_generics #where_clause {
+            type Buffers = #buffer_struct_ident #ty_generics;
+        }
+
+        #buffer_struct
+
+        impl #impl_generics #struct_ident #ty_generics #where_clause {
+            fn select_buffers(
+                builder: &mut ::bevy_impulse::Builder,
```
mxgrey, line 38: I'd rather not have `builder` as an argument. I think we should leave it to users to call `output.into_buffer(builder)` from outside of `select_buffers` and have `select_buffers` just take in the specific buffer types that are expected. This will especially matter for supporting specialized buffer types like `AnyBuffer` and `JsonBuffer`.

koonpeng:
> No, this is not the case. AnyBuffer can be pulled from to produce a AnyMessageBox and JsonBuffer can be pulled from to produce a JsonMessage. Both have implemented the Joined trait so you can use Joined::pull on them just like any concrete buffer type.

I thought the idea is to auto downcast `AnyBuffer`, but if that is not the case, then it would actually solve many of the invalid code generation. Though I think it would also be confusing the difference between `AnyBuffer` and `Buffer<AnyMessageBox>`.

> I think this should be easy to fix by adding a field along the lines of _ignore: ::std::marker::PhantomData<fn( #impl_generics )> into the _Buffers struct. It might be a little tricky to strip the trait bounds out of #impl_generics, but it should be doable by removing all substrings that start with : and end with either , or >.

I didn't think of this, yeah it seems to be a good catch-all solution.

mxgrey:
> I thought the idea is to auto downcast AnyBuffer,

It's actually the opposite. The idea is if a user wants to **_obtain_** an `AnyMessageBox` in their `JoinedValue` struct, they should be able to receive that `AnyMessageBox` from **_any_** kind of buffer by **_upcasting_** the buffer into an `AnyBuffer`.

Like you said, trying to pull a concrete message from an `AnyBuffer` would require us to know how to convert it into the concrete message type, which can't always be done.

> Though I think it would also be confusing the difference between `AnyBuffer` and `Buffer<AnyMessageBox>`.

I expect this won't be a very common use case, and only advanced users would even be interested in obtaining an `AnyMessageBox`.

I do think the `JsonMessage` / `JsonBuffer` use case will be more common, so we'll still need to deal with the problem of distinguishing a `Buffer<JsonMessage>` from a `JsonBuffer`. I'll think about how to give clear examples to illustrate the difference in the `JsonBuffer` documentation.

src/buffer/buffer_map.rs
```diff
@@ -373,33 +382,99 @@ mod tests {
...
+            let buffer_i64 =
+                JsonBuffer::from(builder.create_buffer::<i64>(BufferSettings::default()));
+            let buffer_f64 =
+                JsonBuffer::from(builder.create_buffer::<f64>(BufferSettings::default()));
+            let buffer_string =
+                JsonBuffer::from(builder.create_buffer::<String>(BufferSettings::default()));
+            let buffer_generic =
+                JsonBuffer::from(builder.create_buffer::<String>(BufferSettings::default()));
+            let buffer_any =
+                JsonBuffer::from(builder.create_buffer::<()>(BufferSettings::default()));
```
mxgrey, line 438: You may want to move this test into the `json_buffer.rs` module since `JsonBuffer` requires the `diagram` feature, but `buffer_map.rs` doesn't.

koonpeng:
There is another problem with different buffer types.

The 2 ways `IncompatibleLayout` can check the buffer is either with `require_message_type` or `require_buffer_type`. The former doesn't work because it always return a `Buffer<T>`, the latter also doesn't work because it cannot convert other buffers to `Buffer<T>`, and since all buffers in buffer map is `AnyBuffer`, it only works if all buffers in the buffer struct is one of the "special" buffer types.

src/buffer/buffer_map.rs
```diff
@@ -315,12 +315,15 @@ impl Accessed for BufferMap {
...
 
-    #[derive(Clone, JoinedValue)]
+    #[derive(JoinedValue)]
     struct TestJoinedValue<T: Send + Sync + 'static + Clone> {
         integer: i64,
         float: f64,
         string: String,
         generic: T,
+        #[buffers(buffer_type = AnyBuffer)]
+        #[allow(unused)]
```
mxgrey, line 325: I think it would make sense to pass a real value through this `AnyBuffer` and test that it comes out correctly on the other side instead of leaving it unused.

mxgrey:
> The 2 ways IncompatibleLayout can check the buffer is either with require_message_type or require_buffer_type. The former doesn't work because it always return a Buffer<T>, the latter also doesn't work because it cannot convert other buffers to Buffer<T>, and since all buffers in buffer map is AnyBuffer, it only works if all buffers in the buffer struct is one of the "special" buffer types.

I don't follow what the issue is here. There's a [reference implementation](https://github.com/open-rmf/bevy_impulse/blob/aa22f87fb645b84bdc1d171c24e91c9c09d528dc/src/buffer/json_buffer.rs#L1385-L1388) of this for a struct that contains a `JsonBuffer`. The key is to use `require_message_type::<T>` in the normal case and then use `require_buffer_type::<SpecialBufferType>` when the user is overriding the buffer type.

I understand that procedural macros are easiest to write when everything follows a very consistent pattern, but I believe it should still be possible to generate a `proc_macro2::TokenStream` that has the correct line for each buffer and then insert that token stream into the function body.

mxgrey:
> Another problem is that AnyMessageBox cannot impl Clone. I dropped it and the tests still works but I don't know how much of a blocker is it for actual usage.

This is to be expected. We can't assume that all message types implement clone, and the uncloneability of some message types is accounted for throughout this library. I think we can save it for a future PR to support a `CloneableMessageBox` type that implements `Any + Clone`.

koonpeng:
> > The 2 ways IncompatibleLayout can check the buffer is either with require_message_type or require_buffer_type. The former doesn't work because it always return a Buffer, the latter also doesn't work because it cannot convert other buffers to Buffer, and since all buffers in buffer map is AnyBuffer, it only works if all buffers in the buffer struct is one of the "special" buffer types.
> 
> I don't follow what the issue is here. There's a [reference implementation](https://github.com/open-rmf/bevy_impulse/blob/aa22f87fb645b84bdc1d171c24e91c9c09d528dc/src/buffer/json_buffer.rs#L1385-L1388) of this for a struct that contains a `JsonBuffer`. The key is to use `require_message_type::<T>` in the normal case and then use `require_buffer_type::<SpecialBufferType>` when the user is overriding the buffer type.
> 
> I understand that procedural macros are easiest to write when everything follows a very consistent pattern, but I believe it should still be possible to generate a `proc_macro2::TokenStream` that has the correct line for each buffer and then insert that token stream into the function body.

We actually can't without using hacks, procedural macros work on tokens and does not have type information. `AnyBuffer` may not be `::bevy_impulse::AnyBuffer`, any ident could also be `::bevy_impulse::AnyBuffer` due to type alias etc, downstream users may also implement their own special buffers, all of these would break the proc macro. imo the only proper way is to have a consistent pattern, I'm not very in depth on the impl of the buffers but I think it should be possible to register a downcast `Fn` when a `Buffer<T>` is converted to `AnyBuffer`? If so then we could allow `require_buffer_type` to be able to downcast back into the original `Buffer<T>`.

koonpeng:
> I'm not very in depth on the impl of the buffers but I think it should be possible to register a downcast Fn when a Buffer<T> is converted to AnyBuffer? If so then we could allow require_buffer_type to be able to downcast back into the original Buffer<T>.

37ee9b948387a45e3c8a1138300252d259b809a2 not sure if this is the right place to add it. But this does fix `require_buffer_type` not able to downcast to `Buffer<T>`.

mxgrey:
> https://github.com/open-rmf/bevy_impulse/commit/37ee9b948387a45e3c8a1138300252d259b809a2 not sure if this is the right place to add it. But this does fix require_buffer_type not able to downcast to Buffer<T>.

I hadn't thought of doing that, but it makes enough sense in order to fully generalize the `.downcast_buffer<B>` method. You picked the right place to put it.

mxgrey:
> downstream users may also implement their own special buffers, all of these would break the proc macro. 

I don't agree with this statement and in fact we **should** support downstream specialized buffers. Everything else in the way that I've implemented `AnyBuffer` accommodates it, and the macro should be able to as well.

mxgrey:
Something else that occurred to me, if we rule out the case of

```rust
#[bevy_impulse(buffer = "AnyBuffer")]
field: T,
```

then we no longer need to worry about introducing an

```rust
_ignore: ::std::marker::PhantomData<fn( #generics )>,
```

field to the `_Buffers` struct since specialized buffer types have to map to a concrete output value, making it impossible to have a specialized buffer that outputs a generic without that generic being part of the specialized buffer signature.

Simply put, I think we can totally forget about the need for `PhantomData`, at least until someone can come up with a concrete and legitimate example where it would be needed.

koonpeng:
> > downstream users may also implement their own special buffers, all of these would break the proc macro.
> 
> I don't agree with this statement and in fact we **should** support downstream specialized buffers. Everything else in the way that I've implemented `AnyBuffer` accommodates it, and the macro should be able to as well.

What I mean is that if we do special case for `AnyBuffer` or `JsonBuffer`, then it won't work for custom special buffers that downstream creates. In order support to *any* special buffers, the generated code must be consistent and must not depend on assumptions and edge cases. Allowing downcast to `Buffer<T>` does allow us to have the generated code work in all cases (as long as the special buffer register the correct downcast handlers).

koonpeng:
> Something else that occurred to me, if we rule out the case of
> 
> ```rust
> #[bevy_impulse(buffer = "AnyBuffer")]
> field: T,
> ```
> 
> then we no longer need to worry about introducing an
> 
> ```rust
> _ignore: ::std::marker::PhantomData<fn( #generics )>,
> ```
> 
> field to the `_Buffers` struct since specialized buffer types have to map to a concrete output value, making it impossible to have a specialized buffer that outputs a generic without that generic being part of the specialized buffer signature.
> 
> Simply put, I think we can totally forget about the need for `PhantomData`, at least until someone can come up with a concrete and legitimate example where it would be needed.

Yeah, in fact the latest code already does exactly that, though I left the code there because this logic of converting from generic tokens to a TypePath of PhantomData could be useful in the future? https://github.com/open-rmf/bevy_impulse/blob/a00f09f5fbc66b0d2c63fd131e625b967e3d6621/macros/src/buffer.rs#L63-L77

mxgrey:
> I left the code there because this logic of converting from generic tokens to a TypePath of PhantomData could be useful in the future? 

Some would say it's bad practice to leave dead code hanging around, but I don't mind letting it linger for a while just in case. I would just recommend making it `::std::marker::PhantomData<fn( #(#lifetimes,)* #(#ty_params,)*)>` instead because the `fn(...)` allows the field to automatically implement `Copy`, `Send`, and `Sync` which for some reason doesn't work if it's a regular tuple.

mxgrey:
> What I mean is that if we do special case for AnyBuffer or JsonBuffer, then it won't work for custom special buffers that downstream creates.

I don't agree with this statement and I'll need you to concretely demonstrate a problem with the approach before I can believe it.

You should be able to take the tokens `JsonBuffer` directly out of the string inside of `#[bevy_impulse(buffer = "JsonBuffer")]` and make it the value of `#buffer_type` inside the token stream of `compatibility.require_buffer_type::< #buffer_type >(#map_key, buffers)`.

Please give that a try and then show me how it's unable to work.

Users will be able to use this exactly as-is with downstream / third-party specialized buffers types.

mxgrey:
> Allowing downcast to Buffer<T> does allow us to have the generated code work in all cases (as long as the special buffer register the correct downcast handlers).

What you're describing here doesn't fit the use case of the buffer specialization, as far as I can figure. Can you demonstrate how your approach would work for [this test case](https://github.com/open-rmf/bevy_impulse/blob/5dd9a301a29516e2b95bf473f490d31d793bee0b/src/buffer/json_buffer.rs#L1356-L1426)?

mxgrey:
>> What I mean is that if we do special case for AnyBuffer or JsonBuffer, then it won't work for custom special buffers that downstream creates.
>
> I don't agree with this statement and I'll need you to concretely demonstrate a problem with the approach before I can believe it.

Actually I just did a review of the current code and it looks like you're already doing exactly what I've been asking for, so I'm now even more confused about why you've been saying it can't work.

mxgrey:
After looking back over our conversation, I think I may have misunderstood what you think of as a "special case".

In my mind the "special case" is that we have to apply the `#[...(buffer = "AnyBuffer")]` attribute to the field that needs a specialized buffer type, so when you said that we can't have special cases I thought you were saying that we can't use attributes for specialized buffers.

I guess what you were actually saying is that we shouldn't have special treatment for internal specialized buffer types, and that's something I fully agree with. I never wanted us to treat `AnyBuffer` or `JsonBuffer` differently than any other special/custom buffer types.

I'll do a full review of the PR now that it looks like everything is being supported the way that I had been hoping for.

koonpeng:
> > What I mean is that if we do special case for AnyBuffer or JsonBuffer, then it won't work for custom special buffers that downstream creates.
> 
> I don't agree with this statement and I'll need you to concretely demonstrate a problem with the approach before I can believe it.
> 
> You should be able to take the tokens `JsonBuffer` directly out of the string inside of `#[bevy_impulse(buffer = "JsonBuffer")]` and make it the value of `#buffer_type` inside the token stream of `compatibility.require_buffer_type::< #buffer_type >(#map_key, buffers)`.
> 
> Please give that a try and then show me how it's unable to work.
> 
> Users will be able to use this exactly as-is with downstream / third-party specialized buffers types.

```rust
type FooBuffer = Buffer<Foo>;

#[derive(JoinedValue)]
struct TestWithCustomBuffer {
  normal_data: String,

  #[buffers(buffer_type = JsonBuffer)]
  json_data: JsonMessage,

  #[buffers(buffer_type = FooBuffer)]
  foo_data: Foo,
}

// somewhere in generated code
fn try_from_buffer_map(buffers: &::bevy_impulse::BufferMap) -> Result<Self, ::bevy_impulse::IncompatibleLayout> {
  // there is no customization of the buffer type, so the buffer type is `Buffer<String>`, `IncompatibleLayout::require_buffer_type`
  // cannot downcast to `Buffer<T>`, so we need to use `IncompatibleLayout::require_message_type`.
  let normal_data = if let Ok(buffer) = compatibility.require_message_type::<String>("normal_data", buffers) {
      buffer
  } else {
      return Err(compatibility);
  };

  // A custom buffer type is given, it is `JsonBuffer` so we assume that we can use `IncompatibleLayout::require_buffer_type`.
  let json_data = if let Ok(buffer) = compatibility.require_buffer_type::<JsonBuffer>("json_data", buffers) {
      buffer
  } else {
      return Err(compatibility);
  };

  // A custom buffer type is given, it is `FooBuffer`, we don't know what `FooBuffer`, but 
  // `IncompatibleLayout::require_buffer_type` can convert between "any" special buffers, so we should be able to use it.
  // However, this fails because `FooBuffer` is actually a `Buffer<Foo>` and we need to use `IncompatibleLayout::require_message_type`.
  let foo_data = if let Ok(buffer) = compatibility.require_buffer_type::<FooBuffer>("foo_data", buffers) {
      buffer
  } else {
      return Err(compatibility);
  };
}
```

The main problem is that `require_buffer_type` works for any type of buffers **except** `Buffer<T>`. If we generate different code based on the buffer type in the proc macro, then we would get situations where we use the wrong function to convert the buffers.

In order to support *any* buffers, I think we need to avoid generating code based on the buffer type, we must consider any `buffer_type` options we receive as opaque and it may even be a `Buffer<T>`. A solution so that we do not have to make assumptions is like serde where there is another option to give it a function that converts AnyBuffer to whatever buffer that is specified in `buffer_type` (and we must not consider `AnyBuffer` or `JsonBuffer` as special in anyway). But if `require_buffer_type` *can* downcast to `Buffer<T>`, then we wouldn't have this problem in the first place and it would work with *any* buffer that downstream defines. An option to customize the conversion function is still useful but imo not as critical now as downstream can ensure they register the correct handlers so their buffers work.

koonpeng:
> After looking back over our conversation, I think I may have misunderstood what you think of as a "special case".
> 
> In my mind the "special case" is that we have to apply the `#[...(buffer = "AnyBuffer")]` attribute to the field that needs a specialized buffer type, so when you said that we can't have special cases I thought you were saying that we can't use attributes for specialized buffers.
> 
> I guess what you were actually saying is that we shouldn't have special treatment for internal specialized buffer types, and that's something I fully agree with. I never wanted us to treat `AnyBuffer` or `JsonBuffer` differently than any other special/custom buffer types.
> 
> I'll do a full review of the PR now that it looks like everything is being supported the way that I had been hoping for.

Yeah, maybe I should have made it more clear. I think that if we treat `AnyBuffer` or `JsonBuffer` (or any other buffers) differently, then we would get a situation where the generated code is wrong due to type alias, downstream custom buffers etc.

mxgrey:
I'm glad this has come out so nicely. It works exactly the way I was hoping it would. The `__bevy_impulse_{}_Buffers` technique for avoiding symbol collisions and/or generally hiding the struct is interesting. I suppose it does make sense to force users to explicitly name the struct if they want to be able to use the struct directly.

My only points of feedback are on some of the attribute names. I realize that attribute naming doesn't have any rigorous objective standards so it's going to be a very subjective topic, but I'd suggest a few tweaks to the current naming:

1. Instead of `#[buffers(...)]` I'd suggest we do `#[join(...)]` or `#[joined(...)]`. The struct that we're tagging is a **joined** value, not a set of buffers, so I think it's better to emphasize that the attributes are related to joining.

2. I'm worried that `buffer_struct_name = "..."` will make it sound like we're naming the struct of _a buffer_ when we're actually naming a struct that will contain a collection of named buffers. I'd suggest changing that attribute name to either `buffers_struct_name = "..."` or `buffer_map_name = "..."`.

3. Instead of `buffer_type = "..."` I think it would be enough to just have `buffer = "..."`. Combined with (1), the user would write e.g. `#[join(buffer = "JsonBuffer")]`. That being said, I don't feel strongly about this one. If you think it should be `#[join(buffer_type = "JsonBuffer")]` I'd be fine with that.

koonpeng:
> I'm glad this has come out so nicely. It works exactly the way I was hoping it would. The `__bevy_impulse_{}_Buffers` technique for avoiding symbol collisions and/or generally hiding the struct is interesting. I suppose it does make sense to force users to explicitly name the struct if they want to be able to use the struct directly.
> 
> My only points of feedback are on some of the attribute names. I realize that attribute naming doesn't have any rigorous objective standards so it's going to be a very subjective topic, but I'd suggest a few tweaks to the current naming:
> 
> 1. Instead of `#[buffers(...)]` I'd suggest we do `#[join(...)]` or `#[joined(...)]`. The struct that we're tagging is a **joined** value, not a set of buffers, so I think it's better to emphasize that the attributes are related to joining.
> 2. I'm worried that `buffer_struct_name = "..."` will make it sound like we're naming the struct of _a buffer_ when we're actually naming a struct that will contain a collection of named buffers. I'd suggest changing that attribute name to either `buffers_struct_name = "..."` or `buffer_map_name = "..."`.
> 3. Instead of `buffer_type = "..."` I think it would be enough to just have `buffer = "..."`. Combined with (1), the user would write e.g. `#[join(buffer = "JsonBuffer")]`. That being said, I don't feel strongly about this one. If you think it should be `#[join(buffer_type = "JsonBuffer")]` I'd be fine with that.

I think most of these makes sense, only thing is that I think we shouldn't use "strings" in the attributes, e.g. `#[join(buffer = "JsonBuffer")]` instead of `#[join(buffer = JsonBuffer)]`. Using the ident directly allows for better compiler checks.

mxgrey:
> we shouldn't use "strings" in the attributes, e.g. `#[join(buffer = "JsonBuffer")]` instead of `#[join(buffer = JsonBuffer)]`.

I agree that what you're suggesting looks nicer and probably allows code completion to work better, but I think the `attribute = "value"` convention exists to help deal with syntax that could create ambiguity for the attribute parser. For example, what if someone has a custom buffer type that takes two generic arguments, e.g.:

```rust
#[joined(buffer = CustomBuffer<A, B>)]
```

When the parser splits the attributes based on the comma, I think it'll get `buffer = CustomBuffer<A` and then `B>` as separate attributes. Whereas if we use quotes then `#[joined(buffer = "CustomBuffer<A, B>")]` will definitely get parsed correctly.

mxgrey:
> what if someone has a custom buffer type that takes two generic arguments

I've tried this out now, and there's actually no parsing issue at all. It seems `syn` is smart enough to know the comma belongs to the `< .. >` section. Tbh I no longer understand why `serde` uses quotes for its attribute values, but it seems that it's not necessary.

I'll push a change that adds a unit test for a buffer with multiple generic arguments.

mxgrey (approved):
In [b110e87](https://github.com/open-rmf/bevy_impulse/pull/53/commits/b110e870eadcd36162aea93b38ee0e6f99b974b4) I added a test to make sure that everything works for buffer types with multiple generic arguments, and in fact it does.

At the same time I fixed how `Clone` is implemented for the generated buffer structs because the existing implementation would not have worked if any generic argument did not have the `Clone` trait, even though buffers can be cloned regardless of any generic arguments.

Also all of our internal buffer types implement `Copy` so I think we should implement `Copy` for the generated buffer structs by default. I added a `noncopy_buffer` attribute e.g. `#[joined(buffer = CustomBufferType, noncopy_buffer)]`, which will disable generating the Copy trait whenever it's applied to at least one field in the struct.

koonpeng:
Thanks for the improvements!

PR #55, opened 2025-02-18
avoid conflict in Promise and Iterator

koonpeng:
Not sure if it is just me but rust-analyzer thinks `Promise` is an `Iterator` and `Promise::take` conflicts with `Iterator::take`. Switched to the full form avoid the conflict.


mxgrey (approved):
Nice fix!

I've been seeing this for a long time but I shrugged it off since it's clearly a bug in rust-analyzer rather than a bug in our code. But this is a great workaround 👍

PR #57, opened 2025-02-20
Rename traits to improve semantics

mxgrey:
I haven't been very satisfied with some of the trait naming, especially around `Buffered` and `Joined`. The `-ed` suggests that the operation has been performed, so I think `Joined` makes more sense to describe the value that emerges from the join operation rather than the buffers that are doing the joining. Instead the name for the traits that get applied to collections of buffers have been renamed to use `-ing`, e.g. `Buffering` to indicate that they are carrying out the action rather than that the action has been completed (`-ed`).

I've made the following renames:
* `Bufffered` -> `Buffering`
* `Joined` -> `Joining`
* `Accessed` -> `Accessing`
* `JoinedValue` -> `Joined`
* `BufferKeyMap` -> `Accessor`

I don't mind iterating on these names. The `BufferKeyMap` -> `Accessor` change is one that I especially feel uncertain about. I don't know if `Accessor` is a clear enough name for the trait, whereas `BufferKeyMap` is a lot more explicit about its intention.


koonpeng (approved):
I don't really mind the naming but just to point out that I haven't seen the -ing form (present continuous?) used to name traits before.

mxgrey:
> I haven't seen the -ing form (present continuous?) used to name traits before.

Yeah, it's probably not super conventional, but I think it makes the most sense grammatically for our particular case since we need to distinguish between `Joined` (an object which has been joined, the output of a join operation) and `Joining` (objects which are responsible for joining data together). We could consider `Joiner` instead of `Joining` but then that would be confusing next to `Accessor` which is the output of an access operation rather than the thing which performs the access operation (`Accessing`).

And as far as trait grammar vs convention goes, `Clone` and `Copy` should really be `Cloneable` and `Copyable` from a grammatical point of view, but they've been simplified to `Clone` and `Copy` for the sake of being terse. So I think the conventions are pretty wishy-washy at the end of the day.

Language is hard.

Issue #59, opened 2025-03-19
Support all operation types in diagrams


mxgrey:
This issue ticket tracks the remaining operations that have not yet been added to the diagram feature.

- [ ] fork option
- [ ] trigger / fork trigger
- [ ] collect
- [ ] spread
- [ ] injection
- [x] scope
- [ ] cleanup
- [ ] trim
- [ ] gate open/close
- [x] cancellation filter
- [x] disposal filter
- [ ] push into buffer (insert a value into a buffer and then emit a trigger)
- [x] cloning from buffers instead of pulling
- [ ] explicit serialize
- [ ] explicit try deseserialize

PR #63, opened 2025-03-23
Redesign how diagram workflows are built

mxgrey:
While writing unit tests for #51 I discovered a critical flaw in the way the diagram workflow builder was designed. Up until now diagrams have not correctly supported loops. More generally, if multiple operations are supposed to feed into a single node, only one of those operations will actually be connected to the node and the rest will be ignored. This means the generated workflow is not connected according to what the diagram prescribes. This same problem exists for all operations that take a single message as input.

Despite being called "singe input" operations, there should never be a restriction on how many different operations in the diagram can connect their output into it. The meaning of "single input" is simply that the operation will be **triggered for each single input message that gets delivered to it**, as opposed to funneling operations like `listen` or `join` which gather input from many different sources and combine them before running. But these single input operations can take in messages from any number of sources, *including feeding their own output back into themselves*.

I couldn't see a clear way to tweak the diagram workflow building implementation to fix this problem, so I wound up rewriting it altogether. Besides fixing the above problem, we get the following benefits from the new design:
* The workflow building is much more modular now, with each piece of the implementation involving fewer lines of code and generally being simpler
* Reduced a lot of implementation cruft that accumulated across the multiple iterations that the code has gone through
* Introducing new operations in the future should be much cleaner and require less boilerplate
* No more unknown diagram errors because we don't need to make as many assumptions
* Implicit serialization and deserialization gets automatically applied to all operations where relevant
* Scope input and output can be any message type, not only limited to `JsonMessage`
* Introduced `on_implicit_error` so users can specify what happens when an implicit operation fails (e.g. implicit serialization or deserialization). They can now choose between cancelling the scope, disposing (ignoring) the error, or passing the error message to any node of their choosing within the scope
* Added a `{ "builtin": "cancel" }` target that allows users to trigger a cancellation of the scope
* It should be feasible to do inference both forward and backward with the new design, although this ability has not been exercised yet
* The new design should allow us to provide an API for downstream users to inject new operations for the diagram builder to support, although this has not been exercised yet


src/builder.rs
```diff
@@ -231,6 +233,75 @@ impl<'w, 's, 'a> Builder<'w, 's, 'a> {
...
+        (
+            InputSlot::new(self.scope, source),
+            T::unzip_output(Output::<T>::new(self.scope, source), self),
+        )
+    }
+
+    /// Create an operation that creates a fork for a [`Result`] input. The value
+    /// inside the [`Result`] will be unpacked and sent down a different branch
+    /// depending on whether it was in the [`Ok`] or [`Err`] variant.
+    pub fn create_fork_result<T, E>(&mut self) -> (InputSlot<Result<T, E>>, ForkResultOutput<T, E>)
```
koonpeng, line 252: I like the addition of these operations which is only accessible with a chain previously, I think this is what I'm missing when I tried to unify the implementation of diagram operations. One thing I think would be good is to reduce the duplicated code between this and in `Chain`, for example, `Chain::fork_result` should now be possible to be implemented as

```rust
    pub fn fork_result<U, V>(
        self,
        build_ok: impl FnOnce(Chain<T>) -> U,
        build_err: impl FnOnce(Chain<E>) -> V,
    ) -> (U, V) {
        let (input, outputs) = self.builder.create_fork_result();
        let u = build_ok(outputs.ok.chain(self.builder));
        let v = build_err(outputs.err.chain(self.builder));
        (u, v)
    }
```

alternatively, this can be implemented by creating a chain and calling

```rust
Chain::fork_result(|chain| chain.output(), |chain| chain.output())
```

src/builder.rs
```diff
@@ -231,6 +233,75 @@ impl<'w, 's, 'a> Builder<'w, 's, 'a> {
...
+    where
+        T: 'static + Send + Sync,
+        E: 'static + Send + Sync,
+    {
+        let source = self.commands.spawn(()).id();
```
koonpeng, line 257: What is the implication of this (if any) in run time? Does creating another entity affect run time performance in any way compared to chaining?

koonpeng (changes requested):
I like these changes, I did try to unify the implementation of diagram operations before, but gave up due to reasons I don't fully remember (something regarding the different between the api of nodes and "chains").

src/diagram/workflow_builder.rs
```diff
@@ -1,456 +1,708 @@
...
+        return Ok(Box::new(ImplicitSerialization::new(input_slot)?));
+    }
+
+    if let Some(deserialization) = ImplicitDeserialization::try_new(input_slot, &registry.messages)?
+    {
+        // The target type is deserializable, so let's apply implicit deserialization
+        // to it.
+        return Ok(Box::new(deserialization));
+    }
+
```
koonpeng, line 602: I think these implicit serde should be added when connecting the targets, adding it here means that operations that use `set_connect_into_target` (which there are none atm) will not be implicitly converted.

src/diagram/workflow_builder.rs
```diff
@@ -1,456 +1,708 @@
...
+    /// This can be used to infer information for building the operation.
+    ///
+    /// The workflow builder will ensure that all outputs targeting this
+    /// operation share the same message type, so getting a single sample is
+    /// usually enough to infer what is needed to build the operation.
+    ///
+    /// During the [`ConnectIntoTarget`] phase this will not return any output
+    /// except for any new output added during the current call to [`ConnectIntoTarget`],
+    /// so this function is generally not useful during that phase.
+    pub fn get_sample_output_into_target(&self, id: &OperationId) -> Option<&DynOutput> {
```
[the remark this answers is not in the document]
mxgrey: I specifically avoided using the word "first" in the name because there's no actual ordering to the inputs. The inputs will show up in the vector in an arbitrary order which could even change between runs of the same diagram, so there's no special meaning to being "first". "Sample" is meant to suggest "this is one example of an input that you will have to connect to". I think "sample" is a better word than "example" for this.

I'm open to tweaking the name of this function, but between "first" and "sample" I would insist on going with "sample". If I were being very pedantic then `get_representative_output_into_target` might be the most accurate name possible, but that seems like a mouthful.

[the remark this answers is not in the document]
koonpeng: how about

`get_any_output_into_target`
`get_one_output_into_target`
`get_output_into_target_type` - returns a `TypeInfo` instead, afaik all uses of this function is only using it for the type info atm, and I don't think we are allowing multiple types into a target. For the case that you need *one* output and ignore the rest (seems like a questionable use case), I think it's reasonable to get all outputs and calling `.first()`.

src/diagram/workflow_builder.rs
```diff
@@ -1,456 +1,708 @@
...
+        &mut self,
+        operation: &OperationId,
+        connect: F,
+    ) -> Result<(), DiagramErrorCode>
+    where
+        F: FnMut(DynOutput, &mut Builder, &mut DiagramContext) -> Result<(), DiagramErrorCode>
+            + 'static,
+    {
+        self.set_connect_into_target(operation, ConnectionCallback(connect))
+    }
```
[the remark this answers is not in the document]
mxgrey: I think the only way to make anything like `impl<F> ConnectIntoTarget for F` work is to add a marker generic parameter to `ConnectIntoTarget`, e.g. `impl<F> ConnectIntoTarget<FunctionMarker> F` which is then incompatible with having a `Box<dyn ConnectIntoTarget>` because we can't exclude the marker parameter from the `dyn ConnectIntoTarget`.

Bevy does some interesting tricks with an `IntoSystem` trait which then produces a `Box<dyn System>`, and we could make that work, but I think it's more convoluted than it's worth when we can just wrap the callback in a struct and be on our way.

I don't mind making `ConnectionCallback` private but I'm not clear on what the value of that is unless we think we might tweak its implementation in the future. Out of any of your recommendations I think I'd prefer adding in the documentation.

src/diagram/fork_clone_schema.rs
```diff
@@ -18,87 +34,103 @@ pub struct ForkCloneSchema {
...
+            .commands
+            .spawn((SingleInputStorage::new(self.id()), UnusedTarget))
+            .id();
+        builder.commands.add(AddBranchToForkClone {
+            source: self.id(),
+            target,
+        });
+
+        DynOutput::new(self.scope, target, self.message_info)
+    }
```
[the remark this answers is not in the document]
mxgrey: `ForkCloneOutput<()>` is not type-erased; it will produce `Output<()>` which is a specific concrete output type of `()` which we often call a "trigger" output. This is fundamentally different from `DynOutput` which can represent any type.

`Box<dyn DynForkCloneOutput>` would work, but I don't love putting `Box<dyn _>` into APIs if it can be avoided. I'll give this some thought.

src/diagram/workflow_builder.rs
```diff
@@ -1,456 +1,708 @@
...
-    pub(super) op: &'a DiagramOperation,
-    pub(super) in_edges: Vec<usize>,
-    pub(super) out_edges: Vec<usize>,
+#[derive(Default)]
+struct DiagramConstruction {
+    connect_into_target: HashMap<NextOperation, Box<dyn ConnectIntoTarget>>,
+    // We use a separate hashmap for OperationId vs BuiltinTarget so we can
+    // efficiently fetch with an &OperationId
+    outputs_to_operation_target: HashMap<OperationId, Vec<DynOutput>>,
+    outputs_to_builtin_target: HashMap<BuiltinTarget, Vec<DynOutput>>,
```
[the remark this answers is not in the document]
mxgrey: There's a specific reason I split out these keys: If they are combined into one map with a `NextOperation` key then we need to clone a string and wrap it inside `NextOperation` every time we want to use `.get(~)` for an `OperationId`.

Theoretically we could change `NextOperation` to `NextOperation<'a>` and use `Cow` inside of it. Or we could introduce some new enum that's exactly the same as `NextOperation` but uses `&'a OperationId` instead of `OperationId`, but neither of these options appeal to me.

> Combining into 1 hash map would eliminate these errors.

We wouldn't be able to get rid of that error type by combining the maps. That error would only occur if a diagram contains `"next": "foo"` but doesn't have a `"foo"` operation defined. Merging these maps wouldn't help that.

src/diagram.rs
```diff
@@ -687,9 +746,20 @@ pub struct Diagram {
...
+    /// serializing and deserializing. These implicit operations may be fallible.
+    ///
+    /// This field indicates how a failed implicit operation should be handled.
+    /// If left unspecified, an implicit error will cause the entire workflow to
+    /// be cancelled.
+    #[serde(default)]
+    on_implicit_error: Option<NextOperation>,
+
+    /// Operations that define the workflow
     ops: HashMap<OperationId, DiagramOperation>,
```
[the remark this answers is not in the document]
mxgrey: My plan is to add a variant to `DiagramOperation` like

```rust
Extension {
    plugin: String,
    config: serde_value::Json,
}
```

but I haven't settled on the API for this, and want to save it for a future PR.

src/builder.rs
mxgrey: There's a specific reason that they don't share an implementation beyond `AddOperation`: The `builder` spawns a new entity for the `InputSlot` whereas `Chain` recycles the previous `Output` as an `InputSlot`. These two different approaches are incompatible without doing some wonky things that will make the implementation more convoluted.

I think funneling the implementation into `commands.add(AddOperation::new( ... ))` is enough code sharing between the two functions. The rest is just spawning inputs/outputs as needed and then formatting the return value into the appropriate structure.

src/builder.rs
mxgrey: Eventually this `InputSlot` will get connected to some `Output`, and at that point the [`try_connect`](https://github.com/open-rmf/bevy_impulse/blob/2d9fe36f0042d2e2ba42e2a5087b19b4c057d3ed/src/builder/connect.rs#L58) command will run, which represents all the overhead for doing it this way. Since it only runs once while the workflow is being built, the runtime overhead is extremely negligible.

src/diagram/workflow_builder.rs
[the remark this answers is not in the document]
mxgrey: Those are some good suggestions. I'll mull over which one makes the most sense to me.

src/diagram/workflow_builder.rs
mxgrey: Correct, that's intentional. I don't want to assume that implicit serialization/deserialization is appropriate 100% of the time, especially if/when we start to support extensions.

With the current API it will be the standard connection behavior if the user simply calls `set_input_for_target`, but they can still use `set_connect_into_target` to define whatever custom connection behavior they want, which might exclude implicit conversions.

I'll add documentation to `set_input_for_target` to indicate that it will automatically provide implicit conversions, and I'll add documentation to `set_connect_into_target` to indicate that it does not support that.

src/builder.rs
koonpeng: Is there any difference when the workflow is running? Does the additional connection cause an additional "hop", resulting in additional copying/moving of messages?

src/diagram/workflow_builder.rs
koonpeng: I thought the idea is to always have implicit conversions for all connections. atm it is just serde, but I think it can be generalized to conversion between any types in the future. But I think always doing implicit conversions is something to consider.

The downside is of course that operations cannot choose to opt out of these conversions, but behavior tree choose this approach as well, as long as a conversion is registered, nodes cannot opt out and the executor will always try to convert if possible[[0]]. I wonder in practice if this implicit conversion ever caused problems for nodes, vs the potential issues that arises when a node forgets to opt in to these conversions.

In any case, I think even if it is decided to make conversions opt in, there should be an api for `ConnectIntoTarget` impls to add implicit conversions, as the logic in `standard_input_connection` could change, and downstream operations would then do implicit conversions differently.

[0]: https://www.behaviortree.dev/docs/3.8/tutorial-basics/tutorial_03_generic_ports

mxgrey: At this layer, anyone who wants to extend the available operations is expected to be a very advanced user. If they have a reason for not using the standard connection behavior provided by `set_connect_into_target` then they may have a reason for not wanting implicit conversions.

The implementation of implicit serialization and deserialization are provided by `ImplicitSerialization` and `ImplicitDeserialization` respectively, which are both public structs. If we support more types of conversion in the future then we'll probably introduce an `ImplicitConversion` struct that covers all the use cases.

We don't currently have an API for downstream users to extend current capabilities, so I'm not looking to perfect that use case right now, as long as we can see that we have a reasonable pathway to introducing it.

src/builder.rs
mxgrey: Nope, the `Connect` command works by replacing all references to the `Output` entity with the `InputSlot` entity, so there will be no difference in the final execution behavior.

src/diagram/workflow_builder.rs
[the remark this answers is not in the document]
mxgrey: I've renamed it `infer_input_type_into_target` and it returns `Option<&TypeInfo>`: [c043fb8](https://github.com/open-rmf/bevy_impulse/pull/63/commits/c043fb8f2548d618752afb61e3a52fdf3e71d4b9)

src/diagram/workflow_builder.rs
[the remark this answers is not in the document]
mxgrey: Added to the docs: [c95d80d](https://github.com/open-rmf/bevy_impulse/pull/63/commits/c95d80df34c1ff055cf42404f181b0d6f7b771b3)

src/diagram/fork_clone_schema.rs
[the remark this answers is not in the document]
mxgrey: I've change the implementation to use a `Box<dyn DynamicClone>` inside of `DynForkCloneOutput`: [03767b5](https://github.com/open-rmf/bevy_impulse/pull/63/commits/03767b5913fc5a80fbd2640461ecedfcd01624ea)

src/diagram/workflow_builder.rs
```diff
@@ -1,456 +1,708 @@
...
+        builder: &mut Builder,
+        ctx: &mut DiagramContext,
+    ) -> Result<BuildStatus, DiagramErrorCode>;
+}
 
-    Ok(Some((input, out_edges)))
+/// This trait is used to connect outputs to their target operations. This trait
+/// will be called for each output produced by [`BuildDiagramOperation`].
+///
+/// It is possible
```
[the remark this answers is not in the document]
mxgrey: I've filled in what I meant to put here: [f8452bd](https://github.com/open-rmf/bevy_impulse/pull/63/commits/f8452bddefc3c95333ea3c3c02bb0259cb8116ba)

src/diagram/workflow_builder.rs
mxgrey: I've elaborated on the behavior provided by each of these methods: [8688ea1](https://github.com/open-rmf/bevy_impulse/pull/63/commits/8688ea1453299b56b0ca822df7821b088c0fd6bc)

koonpeng: I beg to differ, I think having advanced users and/or use case does not precludes having a good api. `ImplicitSerialization` and `ImplicitDeserialization` does contain *most* of the impl, but as can be seen in `standard_input_connection`
https://github.com/open-rmf/bevy_impulse/blob/8688ea1453299b56b0ca822df7821b088c0fd6bc/src/diagram/workflow_builder.rs#L621-L623
using them involves additional checks. I think there is no reason to expect "advanced users" to read the source code to understand how to use it correctly. The way `standard_input_connection` use the 2 impls may also change, I think it is also not reasonable to expect users to be aware of that and update their operations.

All of the default operations does implicit conversions, so I think it is reasonable to say that new operations should allow it unless there is a reason not to, so I think a better api is to have it be opt out instead of opt in. Having it opt in would only increase boilerplate for operations, also I think there *will* be "advanced users" who does not add the conversions not because they intentionally do not want implicit conversions, but because they didn't know, or they may add it wrongly.

Having implicit conversions be optional also cause problems for frontend, each operation must indicate if they support implicit conversion so that the frontend can verify if a diagram is valid. Actually, since the current api have implicit conversion be implementation defined, every operation may have a different "implicit conversion" and must indicate the set of input types they can accept.

mxgrey: Implicit conversions are already opt-out by the fact that the implementer chooses not to use `set_input_for_target` and instead chooses to make their own custom `ConnectIntoTarget` implementation, which I think is going to be a very rare situation.

Fundamentally it is not possible for us to provide an API that simultaneously supports:
1. allowing users to define how connections work in any way that suits the details of their operation, and
2. hijacking their connection implementation to inject implicit conversions

Inside of their `ConnectIntoTarget` implementation they can use the `&mut Builder` to create any kind of operation connections that they want, and those connections are totally opaque to the diagram builder, so it would be impossible to inject conversions without doing a post-hoc analysis of the fully constructed workflow connections that they build (this analysis needs to happen _after_ the `Commands` inside of `Builder` get flushed into the `World`) and then we would need to rewire the connections of the fully built workflow.

The best I can offer is making `standard_input_connection` a public function: [d632e85](https://github.com/open-rmf/bevy_impulse/pull/63/commits/d632e8576f2e5e9625150f8fd8f81434f5adb7d0)

I expect when we support fully general `A` to `B` type conversions, we'll introduce some new struct for general conversions, and/or incorporate that into the implementation of the `standard_input_connection` function.

mxgrey: I'll also point out that this is a premature concern since we don't support operation extensions yet, and I don't expect to support them in the very near future.

I expect the vast majority of "custom operations" will be introduced by users defining libraries of [sections](https://github.com/open-rmf/bevy_impulse/pull/58) rather than by defining new operation primitives. It would only be extremely niche or advanced use cases that someone would feel a need to introduce their own operation primitives from downstream.

mxgrey:
@koonpeng Since this PR is a blocker for buffers, sections, streams, and #59 in general I'm going to move forward with merging this and then get #51 merged and go from there.

If you have any lingering concerns feel free to open issue tickets for them.

PR #73, opened 2025-05-15
fix wasm build

koonpeng:
### Fixed bug

Fix wasm32-unknown-unknown build according to https://docs.rs/getrandom/0.3.3/getrandom/#webassembly-support.
Note that this is not a full fix, as of `getrandom@0.3.3`, it needs a rustflag which needs to be specified by the "end target" and all crates must use the same flags, afaik we cannot specify a rustflag for a particular crate.

### Fix applied

**If your pull request is a bug fix, delete this line and all the below lines.**

**If your pull request is a feature implementation, delete this line and all the above lines.**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ semver = { version = "1.0.24", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 uuid = { version = "1.13.1", default-features = false, features = ["js"] }
+getrandom = { version = "0.3.3", features = ["wasm_js"] }
 
 [features]
 single_threaded_async = ["dep:async-task"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ tracing = "0.1.41"
 strum = { version = "0.26.3", optional = true, features = ["derive"] }
 semver = { version = "1.0.24", optional = true }
 
-[target.'cfg(target_arch = "wasm32")'.dependencies]
+[target.wasm32-unknown-unknown.dependencies]
 uuid = { version = "1.13.1", default-features = false, features = ["js"] }
 getrandom = { version = "0.3.3", features = ["wasm_js"] }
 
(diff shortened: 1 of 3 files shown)


mxgrey:
I think if we combine this with https://github.com/open-rmf/bevy_impulse/pull/74 we should get the CI fully green 🥲

Cargo.toml
```diff
@@ -56,6 +56,7 @@ semver = { version = "1.0.24", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 uuid = { version = "1.13.1", default-features = false, features = ["js"] }
+getrandom = { version = "0.3.3", features = ["wasm_js"] }
```
mxgrey, line 59: I'm a bit concerned about this warning:

> WARNING: It is highly recommended to enable the wasm_js feature only for binary crates and tests, i.e. avoid unconditionally enabling it in library crates.

I tried changing this to only apply for `dev-dependencies`, but that didn't work. The compilation failed while the library was building, before it ever got around to building the tests.

I suppose we're not enabling this feature **_unconditionally_**; we're only enabling it for wasm32 architectures. However, I'm not sure if there are situations where someone might target a wasm32 architecture and not want this feature enabled.

mxgrey: > We could also not change the deps but just hack it so CI passes.

My worry at that point is downstream users will have to go through the trouble of figuring all of this out. In fact it's not clear to me whether builds of downstream crates will be able to inherit the `rustflags` from `.cargo/config.toml`. I guess we'll need to test that out later. For now we can just move forward with this PR to unblock CI.

mxgrey (approved):
I'll need to follow up on this by testing whether downstream packages will still work (in particular the site editor), but for now we can merge this to fix CI.

Cargo.toml
koonpeng: One way is to use a custom build script that sets the cfg, but I think there are reasons why `getrandom` does not do that.

PR #75, opened 2025-05-16
fix section schema

koonpeng:
### Fixed bug

schemars generates schemas with `additionalProperties: false` for enums.
When the enum is flatten, that `additionalProperties: false` is inherited by the parent
struct, which leads to schemas that can never be valid.

Example:

```json
{
  "description": "Connect the request to a registered section.\n\n``` # bevy_impulse::Diagram::from_json_str(r#\" { \"version\": \"0.1.0\", \"start\": \"section_op\", \"ops\": { \"section_op\": { \"type\": \"section\", \"builder\": \"my_section_builder\", \"connect\": { \"my_section_output\": { \"builtin\": \"terminate\" } } } } } # \"#)?; # Ok::<_, serde_json::Error>(()) ```\n\nCustom sections can also be created via templates ``` # bevy_impulse::Diagram::from_json_str(r#\" { \"version\": \"0.1.0\", \"templates\": { \"my_template\": { \"inputs\": [\"section_input\"], \"outputs\": [\"section_output\"], \"buffers\": [], \"ops\": { \"section_input\": { \"type\": \"node\", \"builder\": \"my_node\", \"next\": \"section_output\" } } } }, \"start\": \"section_op\", \"ops\": { \"section_op\": { \"type\": \"section\", \"template\": \"my_template\", \"connect\": { \"section_output\": { \"builtin\": \"terminate\" } } } } } # \"#)?; # Ok::<_, serde_json::Error>(()) ```",
  "type": "object",
  "oneOf": [
    {
      "type": "object",
      "required": [
        "builder"
      ],
      "properties": {
        "builder": {
          "type": "string"
        }
      },
      "additionalProperties": false
    },
    {
      "type": "object",
      "required": [
        "template"
      ],
      "properties": {
        "template": {
          "type": "string"
        }
      },
      "additionalProperties": false
    }
  ],
  "required": [
    "type"
  ],
  "properties": {
    "config": {
      "default": null
    },
    "connect": {
      "default": {},
      "type": "object",
      "additionalProperties": {
        "$ref": "#/definitions/NextOperation"
      }
    },
    "type": {
      "type": "string",
      "enum": [
        "section"
      ]
    }
  }
},
```

Here the section schema needs to have a `builder` or `template` with no additional properties.
Which includes other properties like `type`, `config` etc, but `type` is also required which
breaks the schema.

### Fix applied

Remove `additionalProperties` from the flattened field. The new `unevaluatedProperties` option in new json schema drafts addresses this issue, but schemars uses draft-07 so we can't use it. We could update the schema to draft-2020-12 but it's better not to make potentially breaking change.

**If your pull request is a bug fix, delete this line and all the below lines.**

**If your pull request is a feature implementation, delete this line and all the above lines.**

### DIFF
--- a/diagram.schema.json
+++ b/diagram.schema.json
@@ -145,8 +145,7 @@
                 "builder": {
                   "type": "string"
                 }
-              },
-              "additionalProperties": false
+              }
             },
             {
               "type": "object",
@@ -157,8 +156,7 @@
                 "template": {
                   "type": "string"
                 }
-              },
-              "additionalProperties": false
+              }
             }
           ],
           "required": [
--- a/src/diagram/section_schema.rs
+++ b/src/diagram/section_schema.rs
@@ -17,7 +17,7 @@
 
 use std::{collections::HashMap, sync::Arc};
 
-use schemars::JsonSchema;
+use schemars::{schema::Schema, JsonSchema};
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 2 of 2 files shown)


mxgrey (approved):
Very cool that `schemars` provides the `schema_with =` attribute. Thanks for fixing this!

PR #90, opened 2025-06-26
add missing feature in `bevy_impulse_derive`

koonpeng:
The feature is required to build `bevy_impulse_derive` independently.


mxgrey:
I'd be curious what the use case is that leads to this problem.

I don't think `bevy_impulse_derive` can be used independently of `bevy_impulse` since all of the macros are for traits that are defined in `bevy_impulse`.

My rough understanding of the `full` feature is that it's meant for libraries which are parsing Rust syntax trees beyond what is needed by normal proc macros, and it's left out of the default feature set because it increases compile times.

PR #93, opened 2025-07-02
remove unused `$key$` field

koonpeng:
The field is a leftover from when https://docs.rs/serde_with/latest/serde_with/struct.KeyValueMap.html is used. Now that we are no longer using it, it results in an extra field in the serialized json.

### DIFF
--- a/registry.schema.json
+++ b/registry.schema.json
@@ -111,9 +111,6 @@
     "NodeRegistration": {
       "type": "object",
       "properties": {
-        "$key$": {
-          "type": "string"
-        },
         "config_schema": {
           "$ref": "#/$defs/Schema"
         },
@@ -128,7 +125,6 @@
         }
       },
       "required": [
-        "$key$",
         "name",
         "request",
         "response",
--- a/src/diagram/registration.rs
+++ b/src/diagram/registration.rs
@@ -37,7 +37,6 @@ use crate::{
 use schemars::{generate::SchemaSettings, json_schema, JsonSchema, Schema, SchemaGenerator};
 use serde::{de::DeserializeOwned, ser::SerializeMap, Deserialize, Serialize};
 use serde_json::json;
-use tracing::debug;
 
 use super::{
     buffer_schema::BufferAccessRequest, fork_clone_schema::PerformForkClone,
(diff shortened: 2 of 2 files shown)


mxgrey:
It occurs to me that we're already using "name" to refer to an operation's unique identity within a workflow, i.e. [`OperationName`](https://github.com/open-rmf/bevy_impulse/blob/9b5936bd89b3e6a7ccf014c7425b5d77c04ce2a1/src/diagram.rs#L82).

What if we change the `name` field to `default_label` to indicate "This is a default label that we recommend placing on nodes created by this builder, but users are free to override it".

Then we can consistently use "name" to mean "This is the unique name of the operation within its workflow".

koonpeng:
> From what I remember of our conversations, the id field is what should go into

The `id` or `$key$` in this case refers to `my_node`. It exist solely to control how the struct is serialized/deserialized. What it does is to tell serde to convert a dict into a vec and put the key as the `id` field. The main use is to allow an operation schema struct to self-identify its own id. A problem used to be that this creates multiple sources of truth for the `id`, but now that we are using `Arc`, I think we can consider adding back this serialization.

> Then we can consistently use "name" to mean "This is the unique name of the operation within its workflow".

I'm of the opinion that "name" should not be used as the name of a primary key field. I think it is fine if users use a guid as the "name". In fact the frontend does exactly that.

mxgrey:
> I think it is fine if users use a guid as the "name". In fact the frontend does exactly that.

I suspect the ambiguity of the word `name` is causing us to miscommunicate about what the intention of the field is, which supports your opinion that we should not be using it as a field name.

Maybe a better field name would be `display_text`. It would be human-readable text that would be displayed on the rendered node to allow a user of the editor to understand at a glance what the purpose and behavior of the node is. If a human user only saw a guid displayed on the node, they would have no idea what it's for unless they somehow manage to memorize each guid.

In terms of the need for the `id` field, I think I understand now that we know the ID of each node because we will see its key within [this map](https://github.com/open-rmf/bevy_impulse/blob/main/src/diagram/registration.rs#L189). If that's the case then I am okay with removing the `id` field.

koonpeng:
> Maybe a better field name for name: would be display_text:. It would be human-readable text that would be displayed on the rendered node to allow a user of the editor to understand at a glance what the purpose and behavior of the node is. If a human user only saw a guid displayed on the node, they would have no idea what it's for unless they somehow manage to memorize each guid.

That wouldn't solve the problem as the name (or display text) is optional.

https://github.com/open-rmf/bevy_impulse/blob/8777631b0424a6dc87fbc43ddcae53d3ae54242f/src/diagram/registration.rs#L1492

It defaults to the id if not provided. I am still of the opinion that the name or display text should be required, but from previous discussion, it is designed to be optional so that users do not have to give their nodes a name of they don't want to.

PR #95, opened 2025-07-04
make registry Send

koonpeng:
Making it `Send` is required to make it usable from outside bevy.


koonpeng:
`Sync` would be good but it is currently blocked by usage of `RefCell` and `SchemaGenerator`. Decoupling `SchemaGenerator` will require a bigger refactor. For now we can still store it in `Arc` by wrapping it in `Mutex`.

PR #99, opened 2025-08-08
Diagram editor

koonpeng:
### Implemented feature

This still needs more testing before I would consider it stable, but I thought this is a good time to merge to main.

Even though the change is huge, most of the changes is in `diagram-editor` so there shouldn't be much conflicts.

### GenAI Use

- [ ] I did not use GenAI

Generated-by: Gemini


mxgrey (changes requested):
I'm very excited for bevy impulse to have a functional visual editor! It looks like ReactFlow was a good choice, and I look forward to seeing how far we can go with it.

I don't think we have to be very strict on the first draft of the editor, but I'm seeing a few bugs in the UI that I think are likely to cause significant confusion for users in its current state. I think we should get those sorted out before merging this to `main` where users may feel invited to play around with it.

I've described the bugs below, and included a video for most of them. The last section is some remarks on non-critical improvements that I think we should follow up on if we don't resolve them within this PR.

### Discrepancies between the displayed state and the actual state

For nodes, I see strange things happen with the config field. In the following video I have two nodes that use the `mul` builder. I assign one of them a config of 10 and the other a config of 100. Clicking back and forth between them, eventually they both show a config of 10, even though I never changed the one that was supposed to be 100:

https://github.com/user-attachments/assets/005ad7c5-a0d8-48b8-b990-edcb41c8667f

I've seen other cases where I change the config of a node and then immediately re-click the node to find that its config did not seem to change at all.

I see similar problems with connection settings. When trying to set up a split operation, the UI doesn't respond to changes in what I select. Clicking away and re-selecting the edge allows it to update somewhat, but you can see that the sequence value never seems to update in the edit box, even though it shows correctly as 1 on the board:

https://github.com/user-attachments/assets/cd99d28c-59e0-45d5-9d70-57ec446b019e

### Confusing Join input

I made a test case that uses the `Join` operation and found that it was giving me the wrong result. It wasn't an error message; an incorrect number was calculated. After inspecting the generated diagram json I noticed that only one buffer was being connected into the join even though the UI showed two connections into it. The problem is both connections were both set to `Index 0`, so one of them was being quietly discarded.

I understand if it's too tricky to automatically assign new connections to a unique index/key, but I would recommend these two changes to help users deal with this:
* Have the connection line display the index/key of the buffer inside the join, similar to how connections coming out of Split display their sequence number.
* Before allowing the workflow to be run or the json to be generated, check that there are no duplicate entries for Join and pop out an error if there are. We should do the same for `Fork Result`.

### Unable to choose Key for Join connection

The UI for setting a string key for Join doesn't work. Without that we won't be able to join messages into structs; we'll only be able to join into arrays and tuples.

https://github.com/user-attachments/assets/09054b1b-f5ce-4086-b45c-5461146226db

## Non-urgent quality of life improvements

There are few things that I think could improve the quality of life, although I don't consider these urgent. We can open follow-up tickets for these like #102.
* For operations like Split and Fork Clone that support multiple outputs, use a different icon for the output slot, maybe something like a small cluster of 3 or 4 circles instead of one. Fork Result could have exactly two output circles since it only supports two outputs.
* I like that buffer "outputs" and join/listen "inputs" have the same color to indicate that they can be connected to each other. I think we should do the same for ordinary inputs and outputs. I don't think it makes sense that ordinary outputs are white and ordinary inputs are orange when they can connect to each other. If we want to distinguish inputs from outputs then maybe output icons can be little downward pointing triangles instead of circles.
* The icons for Fork Result and Split should be flipped upside-down. I would also recommend using a question mark in some way for Fork Result as a reference to its similarity to the Rust question mark operator.

mxgrey (changes requested):
I've tried out the editor with the recent fixes, and the behaviors are really good now!

> node operations and node builders do not report the streams they output so we must assume all node operations may output a stream.

You're right, this is something I forgot to follow up on when I added the ability for diagrams to support streams. I'll try to get streams into the node builder registry ASAP.

> But even if we know which node operations may output streams, we can't display it on the UI atm as the edge handles are fixed based on the type of operation atm.

It's not urgent but I think this is worth following up on in a future PR. I'll open an issue ticket for this.

While expanding on the calculator example, I added a fibonacci node builder that streams out numbers of the Fibonacci sequence. It seems the UI does not allow you to connect a stream to an input slot:

https://github.com/user-attachments/assets/6b3bf24a-c70b-499a-984f-031b46639462

Furthermore if I create the diagram manually with the intended `"stream_out"` setup and then upload it to the editor, the UI displays it as connected to the node output:

Screenshot from 2025-08-19 19-09-36

When I look at what kind of diagram gets generated from this, the `print` operation is now connected to the output of `fibonacci` instead of connected to its stream.

I would suggest we do one of the following before merging:
* Since there are still some loose ends with streams (i.e. the stream keys aren't being registered), we could just remove stream support from the diagram editor and leave it for a follow-up.
* If it's not too difficult to fix the current bugs, the first draft of support for streams could allow users to manually type in the key string of each connection coming out of a stream output. The user would have to know the names of the stream outputs, but at least a roundtrip of uploading a diagram and then re-exporting it would work correctly.

koonpeng:
I mistakenly thought that stream outputs and only connect to stream outs. It should be fixed now.

mxgrey (approved):
This is an excellent first release for the diagram editor! I think any remaining follow-up items can be addressed in incremental PRs.

PR #113, opened 2025-09-11
Add support for wasm backend

koonpeng:
Add support for wasm backend which can be used in place of the default rest backend. The wasm backend is "serverless", all the rest api calls are replaced by wasm calls and happens on the browser directly.

Unlike the rest backend, which embedded the frontend payload into the server, the wasm backend requires both the wasm payload and frontend to be built from source. There is a [README](https://github.com/open-rmf/bevy_impulse/blob/koonpeng/calculator-wasm/diagram-editor/wasm/README.md) that describes how to do it.

A demo is pushed to `gh-pages` branch, but I don't have the permissions to enable github pages. As a test, I hosted a demo on my own repo https://koonpeng.github.io/bevy_impulse_diagram_editor_demo/.

### Caveat

The current wasm implementation manually polls promise until it is finished
 https://github.com/open-rmf/bevy_impulse/blob/6da83739b7f37f7ef03604722f7b6a786bbae3c4/diagram-editor/wasm/src/executor.rs#L46-L55
This does not block the frontend but the bevy app only updates when there is a response pending. Continuous service and other non bevy systems will not run after the response is finished.
 
 In order to keep the systems running, we would need to run the bevy_app in a service worker. It does not impact the calculator demo so we can put that as a separate feature to be implemented subsequent prs.

> [!NOTE]
> This includes #112.


mxgrey (approved):
These UI improvements are fantastic, and the live web-hosted demo is going to be incredibly useful for introducing the library to people!

Great job with this PR!

mxgrey (approved):

PR #115, opened 2025-09-29
Add help text

### LINKED ISSUE #102: Help text for node and section builders
While experimenting with #99 I found that it's difficult to know what to fill in for the config when setting up a node builder, or to know what the intended behavior of the node is.

Eventually we could probably use the config schema information to provide a more structured UI for filling in the config, but in the meantime it would be good if node and section builders could have a `help_text` field that gives a human-readable description of what the node's behavior is and how to fill in the config box.

koonpeng:
fixes #102.

### DIFF
--- a/examples/diagram/calculator_ops_catalog/src/lib.rs
+++ b/examples/diagram/calculator_ops_catalog/src/lib.rs
@@ -138,7 +138,9 @@ pub struct FibonacciStream {
 
 pub fn register(registry: &mut DiagramElementRegistry) {
     registry.register_node_builder(
-        NodeBuilderOptions::new("add").with_default_display_text("Add"),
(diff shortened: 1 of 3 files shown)


mxgrey:
The intent behind #102 is for the help text to be visible from inside the diagram editor. It looks like this PR doesn't provide that yet.

I imagine we could have something like an `( i )` icon in the node configuration widget that shows the help text when hovered.

Would it be feasible to use something like [react-markdown](https://github.com/remarkjs/react-markdown) to support markdown in the help text? That would be coherent with rustdoc's use of markdown, and I think it would be helpful for making the documentation as readable as possible.

koonpeng:
I was going to submit the frontend changes in another PR, or do you prefer if everything is in this PR?

PR #116, opened 2025-09-29
Fix auto layout crossing edges

koonpeng:
The previous fix for this was in fact wrong, it reorganizes the nodes but in fact, we need to reorganize the edges. Also fix the same crossing issue with fork result.


koonpeng:
After trying many combinations, I find no patterns that can ensure the correct order in all cases. Might be easier to use a more powerful library than hacking dagre to workaround this.

> As a separate issue (unrelated to this PR), if you try to [load the diagram](https://open-rmf.github.io/bevy_impulse/?diagram=lVFLcsMgDL0L02Uw%2FmDj%2BDIZDCKmMZ8BnDaT8d0rt4tukx086X0kPclHVgs4SSaylBLzxFiSX9XVlmWbtwxJBV%2FAl0oFx0IET5MzbIb742Jd3NYMLIHJbAGpM3PSeqatvCbpqj%2Fh6jMHT07kDilbfE2krpqqRqSAi6sskMn03E8kF5kKlltljNKjoB0HTflwFnQULVDVK6maVnIxamSHePBe68a%2B8oiA4iak2yVB3taCGpASYlr2%2FaiHmgqoa8pr1dC5b2pquqFpDXAOwxE33A6%2B5MbMQlHVqZHy89DTcdRnKhqjZdeCaI0iOMxLmv%2BxfNCAFvNmVw1HJvx4%2BMZ1PH%2FBYo%2FFaZtjyEB2NHgpyLsGBZKzHk%2BCFvv%2BAw%3D%3D) instead of building it from scratch, you'll get both connections coming out of ok:

fixed

mxgrey:
In fact it might be deterministic in alphabetical order by the node ID. Testing a bunch of single-letter ID values, it seems the node whose ID is a lower letter in the alphabet is always placed to the left. This would make sense if the generated JSON is sorting the `ops` keys alphabetically and then the nodes get generated in that order when loading.

In any case, at least the ok/err connection bug on loading is fixed.

mxgrey (approved):

PR #135, opened 2025-10-15
Add option to clone buffers for join operations

koonpeng:
Also add option to serialize joins and hide serialized_join from add operations menu.


mxgrey (changes requested):
Thanks for getting this done so quickly. I ran some tests and found quite a few issues, some of which I was able to fix and others I think it would be better for you to take care of. Some of these problems are related to the Split operation, which I used to create tests for the new join-by-clone feature.

Things I fixed:
* The `sequential` property was being populated with `null` for elements that are supposed to be disposed. The intended behavior would be to put `{ "builtin": "dispose" }` for those fields, but I didn't see a clear way to do it, so I decided to change the schema upstream and allow the schema to accept `null` values, which will simply be treated as disposals.
* The split operation was sending off an `(identifier, value)` pair instead of just a value, but I think the vast majority of users will only want the value. I've changed the semantics for this to just send the value, but we can consider options for toggling this behavior in the future.
* The `clone` field of `Join` was being populated with the buffer element IDs, but it's really meant to take the identifier of the buffer within the join itself. This has been fixed.

Still needs fixing:
- [x] Even though the diagram editor will export diagrams with `null` in the elements of `sequential` for Split, it will not load such diagrams, causing round trips to fail.
- [x] Even though the `clone` field of Join is exported correctly, it does not get loaded correctly when importing a diagram.

PR #137, opened 2025-10-17
fix packaging diagram-editor

koonpeng:
cargo excludes gitignore by default, so we have to either commit the dist tarball or use flags to disable the check every time we want to publish. This used the former, with an update to the build script to ensure that the tarball is always up to date.

Tested with an external package depending on the tarball created by `cargo package`.


mxgrey (approved):
Thanks for these fixes! I'll attempt to do a crates.io release of the diagram editor and let you know how it goes.

As a nice bonus, before these changes I noticed that cargo was doing a small ~10 second rebuild every time I would run a downstream application that used this diagram editor library, even if there were no code changes anywhere. It was probably a false positive detection of changes in `dist.tar.gz`. After this PR that no longer happens, and downstream applications can just run immediately without any building :tada:

mxgrey:
Release successful! https://crates.io/crates/crossflow-diagram-editor

All I had to do was reduce the keywords down to five: https://github.com/open-rmf/crossflow/commit/0c0c2b49e4d84c0b03bc45c1287a04a4e2ceda81

PR #151, opened 2025-12-16
[Diagram Editor] fix nodes being able to connect across scopes

### LINKED ISSUE #147: [Diagram Editor] Bug: The editor allows connecting nodes from different scopes
Currently the editor allows the user to connect an operation inside a scope to an operation outside of that scope.

Image

This ends up generating an invalid diagram. The error will be caught when the executor attempts to build the workflow, but it might not be visually clear to the user what the error is or how to fix it. Ideally the UI should prevent this kind of connection and notify the user of why the connection was refused.

koonpeng:
fixes #147


mxgrey:
This fixed being able to connect from inside of a scope to outside of a scope, but there is still an issue of being able to connect from outside to inside:

Screenshot from 2025-12-17 13-51-02

koonpeng:
fixed

PR #155, opened 2025-12-24
Diagram description and example inputs

aaronchongth:
### Implemented feature

Fixes https://github.com/open-rmf/crossflow/issues/153.

* added fields `description` and `example_inputs` into `Diagram`
* propagate usage into diagram editor
* added those fields into example diagram json files we host
* new command panel button for `Diagram properties`, which opens a drawer
* Drawer hosts
  * editable diagram description
  * diagram example inputs, with description, can be deleted, copied into clipboard, or executed directly (just opens the execution dialog with the input in the request, users will still need to click the run button)

https://github.com/user-attachments/assets/c796d474-368c-4503-b142-d324960d3086

### GenAI Use

- [ ] I used a GenAI tool in this PR.
- [x] I did not use GenAI

Generated-by:


src/diagram.rs
```diff
@@ -409,6 +409,12 @@ pub struct Diagram {
 
     #[serde(flatten)]
     pub extensions: Option<ExtensionSettings>,
+
+    /// Optional text to describe the workflow.
+    pub description: Option<String>,
+
+    /// Examples of inputs that can be used with this workflow.
+    pub example_inputs: Option<Vec<String>>,
```
mxgrey: A few concerns here:

* For each example input, we should also allow a description, so users can have some context about the input.
* The data type of the input should not be `String`, it should be `JsonMessage`. The `String` type is too ambiguous, i.e. should the value be interpreted as a literal character string or as the text of a JSON message?
* There's no need to wrap `Option` around `Vec`. We can tag this field with `#[serde(default, skip_serializing_if = "is_default")]`. That way the field can just be skipped for empty vectors.

Overall we should see something like

```rust
pub struct ExampleInput {
    pub value: JsonMessage,
    pub description: String,
}
```

```rust
#[serde(default, skip_serializing_if = "is_default")]
example_inputs: Vec<ExampleInput>,
```

src/diagram.rs
```diff
@@ -409,6 +409,12 @@ pub struct Diagram {
 
     #[serde(flatten)]
     pub extensions: Option<ExtensionSettings>,
+
+    /// Optional text to describe the workflow.
+    pub description: Option<String>,
```
mxgrey: We should tag this field with

```rust
#[serde(default, skip_serializing_if = "is_default")]
```

so it can truly be an optional field.

mxgrey (changes requested):
I think this is a good start, but I see some improvements that can be made before we merge it. I've left some inline comments about improving the data structures in the Diagram schema. Besides that, I have some recommendations on improving the UI:

### Editing descriptions

I think we should address the ability to edit and export the descriptions and input examples within this PR, because I believe doing that in a follow-up PR will effectively become a rewrite of all the typescript introduced in this PR.

### Reduce clutter

I don't think it's necessary for the diagram description or example inputs to always be visible. For example, the node builder descriptions are hidden behind an `(i)` icon, and show up in a tooltip:

Screenshot from 2025-12-29 13-40-29

I think we should do something similar for the workflow descriptions. Maybe we can have a `(?)` or `(i)` toggle button on the right-hand side of the `Run Workflow` row that would expand a panel to the right of the Run Workflow widget. That expanded panel would contain a read-only text box of the workflow description, as well as a 📝 button to switch the description text box into an edit mode.

Then for the input examples, we could put a `(?)` or `(i)` toggle button on the right-hand side of the `Request:` row that would expand a panel that lists out the input examples. This would also have a 📝 button to switch into edit mode. This panel would be mutually exclusive with the description panel, so toggling on one would automatically toggle off the other.

These are just rough ideas. I'm happy to further discuss the feasibility or effectiveness of this approach.

PR #169, opened 2026-01-22
Fix example inputs

aaronchongth:
### Fixed bug

* Fixes https://github.com/open-rmf/crossflow/issues/168
* tiny typo fix

https://github.com/user-attachments/assets/03c2dfed-1c1a-4341-a82d-75c2223753da


mxgrey (approved):
Thanks for fixing this so quickly!

PR #179, opened 2026-03-10
Fix: Prevent duplicate buffers in join operations 

### LINKED ISSUE #61: Add check to prevent the same buffer appearing more than once in a single join
Right now it is technically possible for a user to specify the same buffer multiple times within one join operation, but this would result in a `Broken` error while the workflow is running.

We should instead check whether the same buffer has been put multiple times into the join while building the join instead of risking a runtime problem. We can introduce a new variation of the `join` methods that return a `Result`, and the existing `join` method just unwraps the new one.

kishore08-07:
### Fixed bug

- This PR fixes [#61](https://github.com/open-rmf/crossflow/issues/61):Previously, specifying the same buffer multiple times in a single join operation was allowed at build time, but caused a confusing runtime error (Broken) during workflow execution.

### Fix applied
- Added a `DuplicateBuffer` error type and a `safe_join` method to check for duplicate buffers at join construction time.
- The `safe_join`  method returns a clear error if duplicates are found, preventing the workflow from being built incorrectly.
- The existing `join`  method now calls `safe_join`  and panics with a clear message if duplicates are present, preserving the previous panic-on-error API.
- Added corresponding methods to `Joinable`, `IterBufferable`, and `Builder` for ergonomic usage.
- Added tests to ensure duplicate buffers are caught at construction time, both for the error case `safe_join` and the panic case `join`.

- This change ensures that duplicate buffer errors are detected early, providing a better user experience and preventing confusing runtime failures.

### GenAI Use

- [ ] I did not use GenAI

Generated-by:
GPT-4.1


mxgrey (commented):
Thanks for the contribution! This is a good match of what I recommended in the original issue ticket. I left a few comments about adding missing inline documentation.

One additional point we should address is [`fn try_join`](https://github.com/open-rmf/crossflow/blob/ea25c901df02c1c1a1752510669a61e96c35c988/src/builder.rs#L322-L328). This method supports dynamic joining, needed for JSON diagram workflows. It already returns an error message if the input layout is incompatible with the type that is being joined into, but currently there is no check to prevent multiple buffers from being used.

We should change the error type of `try_join` to an enum with variants of `IncompatibleLayout` and `DuplicateBuffer`, and we should look through the `BufferMap` for buffers with duplicate IDs. If there are any duplicate IDs then `try_join` should return the `DuplicateBuffer` variant.

kishore08-07:
Hi @mxgrey , Could you please let me know if any further changes are needed from my side, or if this is good to merge? Thanks!

Issue #172, opened 2026-02-13
[Diagram Editor] Provide hints to users about compatible operations


mxgrey:
Currently the diagram editor allow users to connect any outputs to any inputs, even if they aren't compatible. It would be good if we can add message type inference to the UI and provide warnings when a user has an incompatible connection, and/or provide hints about which connections are compatible.

PR #182, opened 2026-03-15
Improve diagram editor connection compatibility hints

ArizmendiWan:
This PR is an initial implementation for #172. It improves drag-time connection feedback and adds a compatibility-aware next-operation flow in the diagram editor.

Current changes include:
- clearer compatible/incompatible handle feedback while dragging
- compatibility detection for drags starting from either source or target handles
- dropping on empty space to open a filtered "Compatible next operations" popup

https://github.com/user-attachments/assets/96ddae83-f6d9-4bbf-8bd5-48bced8c767e

This PR also includes a small incidental frontend sync fix for a stale generated type/schema reference that was needed to keep the frontend building cleanly.

### GenAI Use
- [√ ] I used a GenAI tool in this PR.
- [   ] I did not use GenAI

Generated-by: Codex Version 26.309.31024 (962) and Claude Code v2.1.76.


diagram-editor/frontend/connection-hint-panel.tsx
```diff
@@ -0,0 +1,91 @@
...
+      tone = 'success';
+      message = `Compatible target: ${connection.toNode.type}`;
+    } else {
+      tone = 'error';
+      message = result.error;
+    }
+  }
+
+  return (
+    <Panel position="top-right">
```
mxgrey: This connection helper is great, but it's often covered up by the Diagram Properties panel that exists on the right side of the canvas.

Let's change this to `top-left` since the left side of the canvas isn't being used for anything.

aaronchongth (changes requested):
Thanks for the contribution. I've just left some high level review comments for now. Will get back to you soon with more a more in-depth review

diagram-editor/frontend/utils/add-operation-catalog.ts
```diff
@@ -0,0 +1,444 @@
...
+  {
+    key: 'sectionOutput',
+    label: 'Section Output',
+    templateOnlyRoot: true,
+    createPreviewNode: () => createSectionOutputNode('preview_output', { x: 0, y: 0 }),
+    createChanges: ({ newNodePosition, nodeManager }) => {
+      const outputId = addUniqueSuffix(
+        'new_output',
+        nodeManager.nodes
+          .filter((n) => n.type === 'sectionOutput')
```
aaronchongth: use helper function https://github.com/open-rmf/crossflow/blob/main/diagram-editor/frontend/nodes/utils.ts#L80

diagram-editor/frontend/add-operation.tsx
```diff
@@ -1,374 +1,163 @@
...
-        }
-      >
-        Section
-      </StyledOperationButton>
-    </ButtonGroup>
+        ))}
+      </ButtonGroup>
+      {operations.length === 0 && (
+        <Typography variant="body2" sx={{ px: 1.5, py: 1.5, width: 260 }}>
+          No compatible operations are available for this output yet.
```
aaronchongth: This is also prompted when filtered via text with no results instead of handling an output, either
* identify which case it is, and provide proper explanations, or
* use more generic wording to cover both cases

I personally would like it to be the first case, since it is more verbose and explanatory to the user

Image

aaronchongth (changes requested):
Thanks for the changes so far, I left a few comments.

Regarding the UI, the buttons should appear below the filter input, instead of on the right.

Image

Image

When there are no valid operations, the text lines up below the filter properly, the buttons should do that too in general

Image

aaronchongth (approved):
Thanks for the added features and validations! The validations and connection restrictions make sense to me, and I didn't manage to trigger any edge cases so far, but I'll leave the final verdict to @mxgrey to see if anything comes to mind that may be an edge case.

Issue #144, opened 2025-11-29
Derive forking for any enum


mxgrey:
Currently there is built-in support for forking Result and Option types, but users may want to fork custom enum types as well. We could provide a `Fork` procedural macro (or maybe `ForkEnum`) that would allow any enum to be forked.

The same macro could also provide an implementation for `Unzippable` for the enums that it's applied to, where the branches of the unzip are conditionally run based on which variant the enum contained. In this case we might consider renaming the `Unzippable` trait to `Forkable`.

PR #184, opened 2026-03-17
Add Fork derive macro for enum forking and unzipping

kishore08-07:
### Implemented feature

This PR implements #144 , It introduces a procedural macro that enables forking and unzipping for any user-defined enum type, extending the existing support for Result and Option.

- Implements a Fork derive macro to auto-generate Unzippable for enums.
- Exposes ForkUnzip operation and related APIs for downstream use.
- Adds runtime tests for enum forking/unzipping.
- Includes concise inline documentation and rustdoc for new APIs and macro logic.
- All changes validated with passing tests.

### GenAI Use

- [ ] I did not use GenAI

Generated-by:
GPT-4.1


mxgrey (changes requested):
Some broad thoughts on this: This PR does a good job of capturing what jotted out in #144 regarding a `Fork` macro that implements the `Unzippable` trait.

However this doesn't quite cover the original intent of allowing any enum to support the semantics that we have for `Result` and `Option`. I think this PR is a good first step in that direction, but I'll probably need to follow up this PR with a more concrete requirements outline for what the `Fork` or `Forkable` trait would look like.

One important distinction I expect between `Unzippable` and `Forkable` is that `Unzippable` provides the Outputs as an anonymous sequence, whereas `Forkable` should refer to specific names. This is especially relevant for Diagrams where users will want to refer to enum variants by name instead of by index numbers, which is what unzip currently uses.

Maybe the best long-term approach is to remove `Unzippable` entirely and just have a `Forkable` trait which uses [`Identifier`](https://github.com/open-rmf/crossflow/blob/4a3dc530c829538d3d830395ad502488f68434bd/src/identifier.rs#L30-L39) to label each of its outputs. We would then phase out the unzip operation from diagrams in favor of a more general `fork` operation that can be used for tuples, Results, Options, all enums, and even structs (by splitting the fields).

I think most of this is beyond the scope of this PR, but I wanted to lay out my thoughts more clearly so we understand the opportunities available for follow-up work.

The one thing that I do think should be changed before merging this is to not allow named fields in a variant of an enum that we're forking. See my inline comment for more detail.

mxgrey:
I've restored the commits of this PR and pushed them to the [`fork_enum`](https://github.com/open-rmf/crossflow/tree/fork_enum) branch in this official repo.

I would like to carefully consider whether to address [this feedback](https://github.com/open-rmf/crossflow/pull/184#pullrequestreview-3973794295) before merging this feature. For now the `fork_enum` branch is available as a jumping-off point if anyone would like to pick it up.

PR #185, opened 2026-03-18
Fix Split edge default and block mixed slot export

kishore08-07:
### Fixed bug
This PR fixes #183  by improving the behavior of Split operations in the Diagram Editor to make them more intuitive, consistent, and error-resistant.

Previously:

- Split edges defaulted to keyed, making edge labeling less discoverable for users.
- The editor allowed mixed output types (keyed and sequence) from the same Split node.
- Export was not proactively blocked for this invalid Split state.

### Fix applied

- Default Split edge type preference now starts with sequence.
- Added Split export validation that rejects mixed keyed and sequence outputs from the same Split source.
- Export button is disabled when mixed Split slot types exist.
- Conflicting Split edges are highlighted in error color until the mismatch is resolved.
- Added regression tests for both behaviors.

[Screencast from 2026-03-18 14-11-01.webm](https://github.com/user-attachments/assets/8bb27ca2-494c-4895-9a92-639671c14cf5)

### GenAI Use

- [ ] I did not use GenAI

Generated-by:
GPT-5.3-Codex


mxgrey:
Note this error that shows up while running the unit tests:

```
 error: crossflow_diagram_editor@0.0.7: Frontend has changed, run `BUILD_FRONTEND=1 cargo build` to update it
```

You'll need to run `BUILD_FRONTEND=1 cargo build` then commit and push the resulting changes.

aaronchongth (changes requested):
Looks like you've reverted all the changes regarding this PR, instead of just reverting the part about the export button, and keeping just the test in the changes.

@kishore08-07 please review your (or the AI's) changes carefully before making any commits. Especially for UI related changes, please always take a quick video of changes that address any feedback specifically. This will help move the review process along.

Please note that the use of AI in this unsupervised manner is a huge security concern and is frowned upon especially in the open source community, where it may have wider implications than expected.

PR #191, opened 2026-05-05
Diagram support for python scripting operation

aaronchongth:
## New feature implementation

* Related to https://github.com/open-rmf/crossflow/issues/167
* Resolves https://github.com/open-rmf/crossflow/issues/199

### Implemented feature

* script environment manager
  * create, edit, delete script environments
  * highlight script nodes using a specific script environment
  * prevents deletion of script environments if script nodes are using the environment
  * script editor (added `react-codemirror` as dependency)
  * reads what script builders are available, also reads language and default script, for now only python is supported
* script node
  * selection of script environment
  * added exporting and connections
* added a context for handling notifications, to provide some feedback when users create, edit or delete script environments. We can migrate error notifications to this, in the future. Keeping this PR easier to review for now
* minor linting

### Discussion points

* only supporting python for now, as each language support is an additional dep to the frontend. It is described in a tooltip for now

https://github.com/user-attachments/assets/be542a13-839f-4641-925a-4f290dca93a1

- [ ] although regex is fun, I can't say for certain the implementation is foolproof. It would be better to use well tested tools to detect functions if possible instead

### GenAI Use

- [ ] I did not use GenAI

Generated-by: Gemini Flash


aaronchongth:
`ci_linux` is not passing because these branches do not have https://github.com/open-rmf/crossflow/pull/196 yet

I've made the following changes according to feedback and our demo,
* https://github.com/open-rmf/crossflow/pull/191/changes/a1e6f32e43bfb1bd6a9c8ee4c781887f663a0e28 - multiple stream outs are supported now, and exporting stream outs from script nodes
* https://github.com/open-rmf/crossflow/pull/191/changes/28a7825d4242937db8e0899d1fff77e2fe29ce5d - add tests for loading diagrams and validating connections (we are using our basic map based validation for now, full inference validation will require https://github.com/open-rmf/crossflow/issues/203), and API level tests where we run the loaded diagrams against their example inputs
* https://github.com/open-rmf/crossflow/pull/191/changes/602867b81683a7f45c9478a86ca092361e77a102 - the forms for script environment builders were abstracted out, with a default form that has the json config style, while developers can implement their own form based on a builder, in this case a form for `process-bound-python` was implemented to allow choosing ownership, defaulting to `script` for the code editor and hiding away the json config

https://github.com/user-attachments/assets/3e6c7633-dcfb-42cc-8328-be754e409236

aaronchongth:
* Rebased onto `python_op_merge`, and tested functionally that the interactions were working
* had to do a bunch of linting and fixes from the previous iteration of this PR as style and ts checking was added on main

https://github.com/user-attachments/assets/c303514d-c469-4fb6-995a-5c5c97c42b1e

mxgrey (approved):
It was a winding road to get here, but I think it's finally ready to land 🎉

PR #192, opened 2026-05-05
Diagram editor testing and linting github worflows

aaronchongth:
### Implemented feature

* added linting and testing github workflows
* fixed linting and biome errors

Most of the fixes were automatically applied by `pnpm exec biome check --write`, with some minor changes and best-practice fixes done by hand.

### GenAI Use

- [ ] I did not use GenAI

Generated-by: Gemini Flash


.github/workflows/ci_frontend.yaml
```diff
@@ -0,0 +1,30 @@
+name: ci_frontend
+
+on:
```
koonpeng, line 3: Add path filtering to run only when needed.

.github/workflows/style.yaml
```diff
@@ -33,3 +33,30 @@ jobs:
 
     - name: clippy
       run: cargo clippy
+
+  frontend-style:
```
koonpeng, line 37: Add path filtering to run only when needed.

koonpeng (approved):

.github/workflows/ci_frontend.yaml
aaronchongth: as discussed offline, since changes to crossflow might break the frontend, we should continue to run this without filtering, until some time in the future we can split it out.

.github/workflows/style.yaml
aaronchongth: as discussed offline, path filtering only works on the workflow level, not job level, at this point it is probably not worth to split out to another workflow

PR #193, opened 2026-05-13
Add default split key

xiyuoh:
Targets https://github.com/open-rmf/crossflow/issues/183 and as discussed during PMC meetings, this PR adds a default split key `unnamed_key` so users know to edit the key string on the edge.


aaronchongth (dismissed):
LGTM, thanks!

xiyuoh:
@aaronchongth ci_linux was failing so I added `BUILD_FRONTEND=1` in [5544938](https://github.com/open-rmf/crossflow/pull/193/commits/5544938c546b3d61c38c94901f85e97e01f1d25a). Not sure if its worth adding it to the other tests but it didn't seem necessary.

.github/workflows/ci_linux.yaml
```diff
@@ -67,7 +67,7 @@ jobs:
     - name: Test all capabilities combined
       run: |
         rm -rf target/
-        cargo test --workspace -F=maximal
+        BUILD_FRONTEND=1 cargo test --workspace -F=maximal
```
aaronchongth: sorry for the back and forth even with these small changes, the error in the CI about `BUILD_FRONTNED` was prompting the user to run `cd diagram-editor && BUILD_FRONTEND=1 cargo build` which will update the binary blob or lockfile, that then needs to be committed as well. We shouldn't have this step in the CI

aaronchongth (changes requested):

aaronchongth (approved):
LGTM, thanks!

Issue #86, opened 2025-06-12
[Feature request]: Introspection and debugging API


koonpeng:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

It would be good if there is a introspection and debugging API to allow

- [ ] Querying the list of currently running nodes
- [x] Query the amount of times a node is activated and the time it took for it to produce an output
- [x] Setting a breakpoint to pause execution when a node is activated
- [ ] Query the source or reason the node is activated
- [x] Resuming a paused node
- [ ] Forcing a node to produce a given output
- [ ] Forcing a node to receive a given input
- [ ] Change the buffers and streams used by a node

### Implementation Considerations

The API should be accessible from outside the workflow, i.e. from a non bevy managed thread. If the API requires any hooks that might impact performance, it should be made optional.

### Alternatives

Detailed logging and tracing for when a node is activated, how long it took, reason for activation etc.

PR #194, opened 2026-05-16
[Diagram Editor] Frontend progress visualization

ArizmendiWan:
Adds the first frontend-visible debugging slice for the diagram editor.

It adds live operation-start visualization in the graph, exposes the existing backend debug WebSocket through the REST API client, and moves Run/Debug controls into the persistent right side panel so runtime status no longer blocks the diagram canvas.

It also adds a small delayed calculator example so the progress visualization is easy to verify manually.

## Changes
- Highlight the currently executing node and retain visited nodes during debug runs.
- Move Run/Debug UI from popover into the right side panel under a `Run` tab.
- Add optional `wsDebugWorkflow` support to the frontend API client.
- Add `DebugSession` handling for backend `operationStarted` and `finish` messages.
- Add `delay_ms` to the calculator example catalog.
- Add `carry_object_progress.json` as a latency-focused progress demo. 

## Demo Video

https://github.com/user-attachments/assets/edce500d-ee40-4edb-b95c-e979093ae12c

### GenAI Use

- [ ] I did not use GenAI

Generated-by: Codex


examples/diagram/calculator/README.md
```diff
@@ -25,3 +32,14 @@ cargo run -- serve
 ```
 
 Then open http://localhost:3000 to run the diagram editor app from your web browser.
+
+To see live progress highlights in the editor, run the server with debugging
+enabled:
+
+```bash
+CI=true cargo run --features crossflow_diagram_editor/debug -- serve --port 3000
```
aaronchongth: was there a reason why `CI=true` is recommended here?

aaronchongth (changes requested):
Thanks for hammering this out! Introspection and debugging will definitely help users more quickly test workflows.

Before I go through the review more thorough, I have some high-level comments about this feature. Some background reference, https://github.com/open-rmf/crossflow/issues/86. Just to clarify, I believe these features regarding tracing events were called/referenced as "debug" , but it is part of introspection tools and should be provided out-of-the-box, and not just in `debug` mode for crossflow. Especially since they provide incredible value to users while incurring very little overhead, I'd suggest that UX-wise, we have the checkbox turned on by default beside the "Run" button in the Run panel, that only toggles the visualization, while all the tracing events are still provided. Let me know what you think or if I had the wrong understanding of this implementation.

I was able to follow the readme and run the example, but it would be cool to also include screen recordings of the new features and how to interact with them on PRs.

examples/diagram/calculator/README.md
ArizmendiWan: > was there a reason why `CI=true` is recommended here?

This flag was used locally to ensure the command behave consistently in non-interactive environments. I removed it in the latest commit.

ArizmendiWan:
> Thanks for hammering this out! Introspection and debugging will definitely help users more quickly test workflows.
> 
> Before I go through the review more thorough, I have some high-level comments about this feature. Some background reference, #86. Just to clarify, I believe these features regarding tracing events were called/referenced as "debug" , but it is part of introspection tools and should be provided out-of-the-box, and not just in `debug` mode for crossflow. Especially since they provide incredible value to users while incurring very little overhead, I'd suggest that UX-wise, we have the checkbox turned on by default beside the "Run" button in the Run panel, that only toggles the visualization, while all the tracing events are still provided. Let me know what you think or if I had the wrong understanding of this implementation.
> 
> I was able to follow the readme and run the example, but it would be cool to also include screen recordings of the new features and how to interact with them on PRs.

I think you’re right about introspection. Progress visualization is broadly useful for monitoring workflows and intuitive understanding of execution, not only for debugging.

I originally approached this from the diagram editor debugging use case, where progress visualization felt necessary to make debugging usable, which is why the current commits use that naming and flow. I agree it would be better to reposition this as a normal introspection/progress feature. I’ll update the PR in that direction.

edit: I just realized that the progress visualization is entirely built on top of debugging backend, which passes the status information of nodes. If we want to make the introspection default, we may have to make debug default too. IMO the overhead is acceptable.

examples/diagram/calculator_ops_catalog/src/lib.rs
```diff
@@ -346,6 +346,27 @@ pub fn register(registry: &mut DiagramElementRegistry) {
         },
     );
 
+    let delay_description = "Wait for a configured number of milliseconds, then \
+        pass the input through unchanged. This is useful for making example \
+        workflow progress visible in the diagram editor debugger.";
+    let delay_examples = [
+        ConfigExample::new("Wait for the default 750 ms.", json!(null)),
```
aaronchongth: Is there a reason 750ms was picked, as opposed to the usual more commonly used 1000ms?

examples/diagram/calculator_ops_catalog/src/lib.rs
```diff
@@ -346,6 +346,27 @@ pub fn register(registry: &mut DiagramElementRegistry) {
         },
     );
 
+    let delay_description = "Wait for a configured number of milliseconds, then \
+        pass the input through unchanged. This is useful for making example \
+        workflow progress visible in the diagram editor debugger.";
```
aaronchongth: Mention in the description that,
* This node is only for testing the progress visualization (also mention the example workflow json), and it should not be used in other workflows to naively introduce delays for async purposes.
* what the default value is, if not configured

diagram-editor/frontend/diagram-side-panel.tsx
```diff
@@ -0,0 +1,404 @@
...
+        >
+          <Tabs
+            value={tab}
+            onChange={(_, value: DiagramSidePanelTab) => onTabChange(value)}
+            sx={{ minHeight: 40 }}
+          >
+            <Tab
+              value="run"
+              label="Run"
+              icon={<MaterialSymbol symbol="play_arrow" />}
```
aaronchongth: I think we should still have the properties tab as the first tab to be displayed by default, this would be very helpful whenever users load a new diagram in, and is able to see the description and example input requests.

diagram-editor/frontend/nodes/base-node.tsx
```diff
@@ -34,7 +40,24 @@ function BaseNode({
     );
 
   return (
-    <Paper>
+    <Paper
+      sx={(theme) => ({
+        outline: debugLatest
+          ? `2px solid ${theme.palette.success.main}`
+          : debugVisited
+            ? `1px solid ${alpha(theme.palette.info.main, 0.35)}`
```
aaronchongth:
```suggestion
            ? `2px solid ${alpha(theme.palette.info.main, 0.35)}`
```

the outline for visited nodes is is same as the default outline, 1px, changing it to 2px will at least make a visible difference

diagram-editor/frontend/nodes/base-node.tsx
```diff
@@ -34,7 +40,24 @@ function BaseNode({
     );
 
   return (
-    <Paper>
+    <Paper
+      sx={(theme) => ({
+        outline: debugLatest
+          ? `2px solid ${theme.palette.success.main}`
```
aaronchongth: so far it looks like we are only display the `success` color on the latest node that has reported it has started. We should have the `success` color on all ongoing active nodes, and only extinguish it back to 2px `info` (visited), once those nodes have reported that it has finished.

Example of this issue, by adding a fork clone in the example json of this PR, and repeating another branch of delays, we see that only the most recent reported delay node is active on each step of the separate branch, when they both should be lit up

https://github.com/user-attachments/assets/890037b6-4c8a-42db-bfed-0a877e7b8032

aaronchongth (changes requested):
Overall, I think the plumbing is working as expected, I've left some comments about the feature as well as the UX.

> edit: I just realized that the progress visualization is entirely built on top of debugging backend, which passes the status information of nodes. If we want to make the introspection default, we may have to make debug default too. IMO the overhead is acceptable.

You're right about that. I also had the discussion with @koonpeng and in the early stages, it was considered to only be used for debugging purposes, but with a greater focus on building and running workflows from the UX perspective, I believe this should all be under the `router` feature. And any mention of "debug" should be renamed to either along the lines of progress visualization or just `wsSession`. Users from the frontend, can choose to enable the visualization or not.

examples/diagram/calculator_ops_catalog/src/lib.rs
ArizmendiWan: There is no specific reason. This is an arbitrary number. I will change the default number to 1000 in the following commit.

examples/diagram/calculator_ops_catalog/src/lib.rs
ArizmendiWan: Notes have been updates in the latest commit

diagram-editor/frontend/diagram-side-panel.tsx
ArizmendiWan: I reordered the side panel tab ordering and set properties tab as default

diagram-editor/frontend/nodes/base-node.tsx
ArizmendiWan: Node effect has been modifed. Now the visited node will show blue glow.

diagram-editor/frontend/nodes/base-node.tsx
ArizmendiWan: Right now the debugging backend only report node start but not finish. Implementing this feature would need to introduce new backend message.

edit: I think we can use MessageSent as the signal of last node finished, so maybe a new message is not needed.

aaronchongth (approved):
LGTM! Thanks for completing this feature, this will definitely help users better understand how their workflows are being executed.

edit: I helped make DCO pass

PR #196, opened 2026-05-16
Pin pnpm version in Linux CI

### LINKED ISSUE #195: Linux CI fails because pnpm latest requires newer Node than runner default
Linux CI currently installs pnpm with `version: latest` before setting up Node.

Recent pnpm latest resolves to pnpm 11.x, which requires Node >= 22.13 and imports `node:sqlite`. The GitHub runner is using Node 20.20.2 at that point, so pnpm crashes before the later `pnpm env use --global lts` step can run.

Likely fixes:
- Pin pnpm to a Node-20-compatible version, or
- Set up Node >= 22 before installing pnpm

ArizmendiWan:
Resolves #195. Pinning pnpm version to 10.31.0, which is compatible with Node 20.


.github/workflows/ci_linux.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
     - name: Setup pnpm
       uses: pnpm/action-setup@v4
```
aaronchongth: https://github.com/pnpm/action-setup is up to version 6 now, perhaps we should use this chance to upgrade it too

.github/workflows/ci_linux.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
...
       with:
-        version: latest
+        version: 10.31.0
 
     - name: Setup nodejs
       run: pnpm env use --global lts
```
aaronchongth: Let's use https://github.com/actions/setup-node for this step too, since we can set up caching with pnpm

aaronchongth (changes requested):
Thanks for the fixes! I left some comments on unchanged lines to hopefully get our CI updated in one go too. I am however unsure if keeping `latest` and `lts` on `v6` will resolve your reported issue, since the version mismatch comes from upstream. If the issue still persists after these updates, we can pin the versions as you have currently already done.

.github/workflows/ci_linux.yaml
ArizmendiWan: Updated in the last commit, and resetted version to latest

.github/workflows/ci_linux.yaml
ArizmendiWan: Updated to setup-node@v4

ArizmendiWan:
> Thanks for the fixes! I left some comments on unchanged lines to hopefully get our CI updated in one go too. I am however unsure if keeping `latest` and `lts` on `v6` will resolve your reported issue, since the version mismatch comes from upstream. If the issue still persists after these updates, we can pin the versions as you have currently already done.

I have updated the pnpm and node setup action in the latest commit. Also updating pnpm allowBuilds setting, as in pnpm 11, onlyBuiltDependencies was removed and replaced by allowBuilds: https://pnpm.io/settings#:~:text=The%20following%20settings%20have%20been%20removed%20in%20v11%20and%20replaced%20by%20allowBuilds%3A%20onlyBuiltDependencies%2C%20onlyBuiltDependenciesFile%2C%20neverBuiltDependencies%2C%20ignoredBuiltDependencies%2C%20and%20ignoreDepScripts.

ArizmendiWan:
Also updated actions version to latest in all CI configs.

aaronchongth (dismissed):
LGTM, thanks for the fix!

edit: apologies #192 was merged with stale workflows (before it got broken), can you help make the changes to those new workflows too? Changes should be similar to #197, which I opened and then realized it was not useful

ArizmendiWan:
> LGTM, thanks for the fix!
> 
> edit: apologies #192 was merged with stale workflows (before it got broken), can you help make the changes to those new workflows too? Changes should be similar to #197, which I opened and then realized it was not useful

The actions in the new workflows are updated too.

aaronchongth (approved):
thanks for the fixes! LGTM

Issue #203, opened 2026-05-29
[Feature request]: Use diagram inference to validate if connections and types are valid


aaronchongth:
### Before proceeding, is there an existing issue or discussion for this?

- [x] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://discourse.openrobotics.org/c/open-rmf).

### Description

At the moment we validate diagram connections and types based on a map that's defined in the diagram editor.

We should instead use the inference system from `inference.rs` to handle validation both diagram connections and data type compatibility.

PR #210, opened 2026-06-20
[Diagram Editor] Enhance compatibility hint with inference

ArizmendiWan:
This PR targets issue #172 and issue #203. 

This PR adds compatibility checks for diagram editor connections while preserving incomplete-diagram editing workflows. The editor now:
- shows green/red handle hints during drag
- explains compatibility results in the connection helper
- filters empty-space add-operation suggestions by compatibility
- allows provisional connections when inference needs more graph context

Compatibility works in two layers:
- **Frontend shape check**
	The editor first rejects impossible connections locally: output-to-output, input-to-input, etc.
- **Rust inference check**
	For a valid-looking candidate, the frontend builds an in-memory diagram with the proposed edge added, then calls `checkCompatibility`. Rust runs editor-only inference on the focused ports for that candidate.  
- Rust returns:
  - `compatible`: the editor may allow the connection.
  - `incompatible`: proven invalid, such as hard compile/layout errors or message type mismatch.
  - `unknown`: API/system failure or unsafe-to-classify error.
  - `provisional`: true when inference needs more graph context, for example buffer/listen/join cases where the type cannot be determined yet.
- If both source and target message types are inferred, Rust compares them using workflow semantics: exact match, registered conversion, JSON serialization/deserialization, or `ScriptMessage` conversion. If inference cannot determine types only because context is missing, the result becomes provisionally compatible, but final compile/run can still reject it later.

https://github.com/user-attachments/assets/99fad0d7-79c1-4f40-8028-9df32360f690

https://github.com/user-attachments/assets/e3509c33-c197-42d6-bc88-ad417a325f51

https://github.com/user-attachments/assets/5d4a892b-3797-498d-abbd-f7d56e19d8e7

### GenAI Use

- [ ] I did not use GenAI

Generated-by: Codex


src/diagram/inference.rs
```diff
@@ -96,6 +96,39 @@ impl Diagram {
...
     ) -> Result<InferredMessageTypes, DiagramError> {
+        let inferences = self.evaluate_message_type_inferences(lookup, boundary)?;
+        Ok(inferences.try_infer_types()?)
+    }
+
+    /// Infer message types only for the requested ports.
+    ///
+    /// This is intended for editor hints where the diagram may be in a
+    /// partially edited state. Unlike [`Self::infer_message_types`], unrelated
+    /// ports that cannot yet be inferred do not make this fail.
```
aaronchongth, line 107: Can you help add a couple of tests for this, to check that the desired ports are referenced, as well as making sure that a partially broken diagram can still pass as long as the valid ports are selected.

diagram-editor/frontend/utils/compatibility.ts
```diff
@@ -0,0 +1,384 @@
...
+  const candidateNodes = cloneJson(nodeManager.nodes);
+  for (const change of nodeChanges) {
+    if (change.type === 'add') {
+      candidateNodes.push(cloneJson(change.item));
+    }
+  }
+
+  const candidateManager = new NodeManager(candidateNodes);
+  const edge = createEdgeFromConnection(connection, candidateManager, edgeId);
+  if ('valid' in edge) {
```
aaronchongth:
```suggestion
  if (typeof edge === 'ValidationError' ) {
```

use a proper type guard instead

diagram-editor/frontend/diagram-editor.tsx
```diff
@@ -841,36 +826,50 @@ function DiagramEditor() {
...
-          }
+          void (async () => {
+            const newEdge = await tryCreateCompatibleEdge(
+              newConnection,
+              oldEdge.id,
+            );
+            if (newEdge) {
+              oldEdge.type = newEdge.type;
+              oldEdge.data = newEdge.data;
+              setEdges((prev) => reconnectEdge(oldEdge, newConnection, prev));
```
aaronchongth:
```suggestion
              const updatedEdge = { ...oldEdge, type: newEdge.type, type: newEdge.data };
              setEdges((prev) => reconnectEdge(updatedEdge, newConnection, prev));
```
we generally try to avoid mutating the provided parameters in a callback, as it may lead to undefined re-rendering behaviors. I'd recommend cloning the object

diagram-editor/frontend/utils/add-operation-catalog.ts
```diff
@@ -458,8 +466,8 @@ export function getVisibleAddOperations(options: {
   });
 }
 
-export function filterCompatibleAddOperations(
-  definitions: AddOperationDefinition[],
+export function filterCompatibleAddOperations<T extends AddOperationCandidate>(
+  definitions: T[],
```
aaronchongth, line 473: I see this change that is related to the changes in `AddOperationCandidate` and `AddOperationDefinition`, but can't seem to see why this change is required. Is some other part of the code base using the `string` based key? Or is the linter complaining?

diagram-editor/server/api/executor.rs
```diff
@@ -116,6 +230,646 @@ pub async fn post_run(
...
+    }
+}
+
+fn can_connect_message_types(
+    registry: &DiagramElementRegistry,
+    source_type: usize,
+    target_type: usize,
+) -> Result<Option<String>, DiagramErrorCode> {
+    if source_type == target_type {
+        return Ok(Some("message types match exactly".to_string()));
```
aaronchongth:
```suggestion
        return Ok(Some("Message types match exactly".to_string()));
```
nitpick, this comment applies to all other Result string too, since we know that this is used at the start of the sentence in the compatibility hints.

aaronchongth (changes requested):
Thanks for handling this feature! I've left some comments, mostly related to the frontend. Overall I've not seen any issues from the usability side,

diagram-editor/frontend/utils/compatibility.ts
ArizmendiWan: Resolved in the latest commit with a type containing a flag for validity.

src/diagram/inference.rs
ArizmendiWan: Resolved in the latest commit with corresponding tests.

diagram-editor/frontend/diagram-editor.tsx
ArizmendiWan: Resolved as suggested.

diagram-editor/server/api/executor.rs
ArizmendiWan: Resolved for all result strings.

diagram-editor/frontend/utils/add-operation-catalog.ts
ArizmendiWan: getRegistryNodeBuilderCandidates creates candidates from registry.nodes with dynamic keys like `node:${builder}` that don't fit the fixed AddOperationKey union. I updated the type to use a template literal (`node:${string}`) instead of a bare string, so the key is now AddOperationKey | RegistryNodeKey

src/diagram/stream_out_schema.rs
```diff
@@ -75,6 +75,12 @@ pub struct StreamOutSchema {
     pub trace_settings: TraceSettings,
 }
 
+impl StreamOutSchema {
+    pub fn name(&self) -> &OperationName {
```
mxgrey: Instead of creating this function, we should actually just change

```rust
pub struct StreamOutSchema {
    pub(super) name: OperationName,
    ...
}
```

to

```rust
pub struct StreamOutSchema {
    pub name: OperationName,
    ...
}
```

The restriction on the visibility comes from a time very early in the development of this library when we weren't sure if we would want these fields to be public. But since we want users to be able to construct and modify these diagrams programmatically, we now expect all fields in the diagram data structure to be public.

aaronchongth (dismissed):
LGTM on the changes for the frontend

Should wait till @mxgrey's review has been addressed before merging

src/diagram/stream_out_schema.rs
ArizmendiWan: This has been resolved in the last commit.

aaronchongth (approved):
LGTM!

PR #211, opened 2026-06-25
Fix: Prevent duplicate buffers in join operations

mxgrey:
This PR is a restoration of #179 since the repo for that PR has been deleted. Luckily I had already fetched most of the commits for that PR on my local machine, so I was able to restore the bulk of the commit history.


mxgrey:
Thanks for the contribution @kishore08-07 , and sorry for how long it took to finish the review. We've had quite a backlog to contend with.

PR #212, opened 2026-07-02
Prevent ESC from discarding edits

ArizmendiWan:
### Fix applied

Prevent `Escape` from closing the script environment manager and node/edge edit popover when focus is inside an active editing field. `Escape` still closes these surfaces when focus is outside text/code editing controls.

### GenAI Use

- [ ] I did not use GenAI

Generated-by: GPT-5.5 with codex


mxgrey (dismissed):
I tested it out, and the behavior is exactly what I hoped for!

mxgrey (approved):

aaronchongth (approved):

PR #216, opened 2026-07-26
[Feature] Allow caching unsaved edits and creating new diagram

ArizmendiWan:
This PR adds tab-scoped draft recovery to the diagram editor so unfinished work can be restored after a refresh.

Current changes include:

- autosave the main diagram, active template graph, diagram properties, templates, source extensions, filename, and unfinished editor buffers
- store one versioned draft per browser tab using pathname-scoped `sessionStorage`
- prompt the user to restore or discard an existing draft on startup
- add a New Diagram action and confirmation before replacing dirty work

### Demo video

https://github.com/user-attachments/assets/6a0a17a6-1424-4285-b912-917ba25beb04

### GenAI Use

We follow [[OSRA's policy on GenAI tools](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)

- [ ] I did not use GenAI

Generated-by: GPT5.6-sol with Codex


diagram-editor/frontend/diagram-editor.tsx
```diff
@@ -839,6 +987,244 @@ function DiagramEditor() {
...
+    };
+  }, [clearStoredDraft, draftContent, hydrationResolved, persistLatestDraft]);
+
+  useDraftPagehideFlush(
+    latestDraftContent,
+    hydrationResolved,
+    setDraftStorageFailed,
+  );
+
+  useBeforeUnloadWarning();
```
aaronchongth: we can probably use `isDirty || hasUncommittedBuffers` to decide on triggering this, maybe pass a boolean into `useBeforeUnloadWarning`?

Can be tested in `use-before-unload-warning.test.tsx` too

diagram-editor/frontend/diagram-editor.tsx
```diff
@@ -839,6 +987,244 @@ function DiagramEditor() {
...
+    autosaveTimer.current = setTimeout(persistLatestDraft, 500);
+    return () => {
+      if (autosaveTimer.current !== null) {
+        clearTimeout(autosaveTimer.current);
+        autosaveTimer.current = null;
+      }
+    };
+  }, [clearStoredDraft, draftContent, hydrationResolved, persistLatestDraft]);
+
+  useDraftPagehideFlush(
```
aaronchongth, line 1082: we can probably use `isDirty || hasUncommittedBuffers` to decide whether to trigger writing the draft workspace, this should prevent the after-refresh prompt to show up if the workspace is clean

can be tested in `use-draft-pagehide-flush.test.tsx` too

aaronchongth (changes requested):
Thanks for putting together this caching mechanism! I've left some UX related comments regarding the interaction when the workspace is clean, but overall it's looking good!

diagram-editor/frontend/diagram-editor.tsx
ArizmendiWan: This has been resolved in the latest commit.

diagram-editor/frontend/diagram-editor.tsx
ArizmendiWan: This has been resolved in the latest commit.

aaronchongth (approved):
Thanks for the fixes. I found that the changes to prevent dialogs from popping out when refreshing new empty diagrams only work the first time the tab is opened, subsequent empty tabs still trigger the prompt. Looking at the current code path, there is no great way to make it handle all the cases as the state will have a cyclical relationship with `resetToNewDiagram()`. 

Overall, I don't think this is a huge issue, and we can come back to this if it ever becomes a requirement (but most likely not)

Thanks for iterating! LGTM!

Issue #126, opened 2025-10-12
[Diagram Editor] Config example browser


mxgrey:
#115 introduces both a help text description of a node builder and a list of config examples with explanations of each example, however the PR only exposes the help text description to the front end, leaving the config examples hidden.

We should follow up with some UI element that allows users to browse and select example configs. This should probably come after #105 since that one is more crucial for UX.

PR #217, opened 2026-08-07
[Diagram Editor] Schema-driven node configuration form

### LINKED ISSUE #105: [Diagram Editor] Allow Config UI to have better support for typing JSON
- [x] Despite appearing to be a multi-line text edit field, the Config UI forces user input to fit in a single line. This makes it difficult to input and edit JSON objects.
- [x] For many node builders, the config is optional, but if you delete the contents of the Config UI after filling something in, it will be reverted to its previous contents after you refresh the edit box. Currently the only way to clear out the contents is to type in `null`. If a user deletes the content of the Config UI, it should remain deleted.
- [ ] It would be great if we can use the config schema information to tell a user when there's an error in their config input. This is relatively low priority since the workflow executor will tell us about the error when we try to run it.

ArizmendiWan:
This PR replaces the free-form JSON config box in the node editor with a form generated from the selected builder's config schema, so a node can be configured without reading the builder's Rust source. This should resolve #105 and #126.

Current changes include:

- generate typed controls (text, number, checkbox-gated `Option`, dropdowns, nested groups) from the builder's `config_schema`
- show the builder description under the Builder field, where it is needed while configuring
- list the builder's `config_examples` in a collapsible section, each applicable in one click
- keep the edit popover on screen when its content grows

### Form behaviour

The form defines the shape of the config: every field it renders has a key behind it, whether or not the user has touched that field. Required keys are filled in when a node is created and when an incomplete config is opened, so a config the form has shown is a config the backend will accept. Optional properties keep their `Set <name>` checkbox.

### Demo Video

https://github.com/user-attachments/assets/7ed70d6e-38ce-4ab2-88b5-2aa615b8c9cc

### GenAI Use

- [ ] I did not use GenAI

Generated-by: Claude Opus 5 with Claude Code


aaronchongth (dismissed):
Awesome feature! I appreciate how we can do this in-memory and all from the metadata, the growing form and node description in the form is a nice touch for UX too. Will need some conflict resolving to do, but generally LGTM

aaronchongth (approved):
Thanks for demo-ing the example of a nested config as well, LGTM

